PR #41, opened 2024-09-18
fix: prevent user from deleting their own roles

Emnaghz:
# Motivation

Fixes # (issue)

## Type of change

# Checklist:


api/src/user/controllers/role.controller.ts
```diff
@@ -142,12 +147,24 @@ export class RoleController extends BaseController<Role, RoleStub> {
...
   @Delete(':id')
   @HttpCode(204)
-  async deleteOne(@Param('id') id: string) {
-    const result = await this.roleService.deleteOne(id);
-    if (result.deletedCount === 0) {
-      this.logger.warn(`Unable to delete Role by id ${id}`);
-      throw new NotFoundException(`Role with ID ${id} not found`);
+  async deleteOne(@Param('id') id: string, @Session() session: ExpressSession) {
+    const roles = (
+      await this.userService.findOneAndPopulate(session.passport?.user?.id, [
```
marrouchi, line 152:
```suggestion
      await this.userService.findOneAndPopulate(req.user, [
```
I think the id can be found in req, can you check ?

api/src/user/controllers/role.controller.ts
```diff
@@ -142,12 +147,24 @@ export class RoleController extends BaseController<Role, RoleStub> {
...
+        'roles',
+      ])
+    ).roles.map((role) => role.id);
```
marrouchi, line 155: Let's break this into multiple instructions
`const currentUser = await ...`
`if (!user) { ...`
`const roles = user.roles.map(....`

marrouchi (changes requested):

api/src/user/controllers/role.controller.ts
Emnaghz: Attempting to access req.user.id results in an error. Instead, the user ID can be accessed via req.session.passport.user.id or session directly like i wrote in my code.

image

medtaher123: it's only a typecheck error, the actual value is there.
You can easily solve that by a simple cast to user but that's not how you should do this in this case.
The user object is already there so there is no need to fetch it again with findOneAndPopulate, You only need to fetch the roles.

But seeing that you didn't use the roles afterwards, you only used their IDs, there is no need to even to fetch the roles, you already have a roles array in the object user containing role ids

Similarly if you did have to fetch the user ( which is not the case), you wouldn't need to populate the roles since you only need the roles ids

api/src/user/controllers/role.controller.ts
```diff
@@ -142,12 +147,29 @@ export class RoleController extends BaseController<Role, RoleStub> {
...
+      try {
+        const result = await this.roleService.deleteOne(id);
+        if (result.deletedCount === 0) {
+          throw new NotFoundException(`Role with ID ${id} not found`);
+        }
+        return result;
+      } catch (error) {
+        throw new NotFoundException(`Role with ID ${id} not found`);
+      }
     }
```
medtaher123, line 173: This code can be made cleaner and more concise:
```suggestion
   if (roles.includes(id)) {
      throw new ForbiddenException("Your account's role can't be deleted");
    }
    const result = await this.roleService.deleteOne(id);
    if (result.deletedCount === 0) {
      throw new NotFoundException(`Role with ID ${id} not found`);
    }
    return result;
   }
```

medtaher123 (changes requested):

api/src/user/controllers/role.controller.ts
medtaher123: You can maybe add a  `this.logger.warn(`Unable to delete Role by id ${id}`);` 
before throwing the error

medtaher123:
I have a suggestion: wouldn't it be better to prevent role deletion only if the role is still in use? In other words, a role can only be deleted after it has been removed from all users who are using it

api/src/user/controllers/role.controller.ts
Emnaghz: Actually the user exists but with an id, first_name and last_name attributes as shown in the following screenshot, so i need to fetch the user data with findOneAndPopulate
image

api/src/user/controllers/role.controller.ts
```diff
@@ -147,12 +153,26 @@ export class RoleController extends BaseController<
...
+    } else {
+      try {
+        const result = await this.roleService.deleteOne(id);
+        if (result.deletedCount === 0) {
+          throw new NotFoundException(`Role with ID ${id} not found`);
+        }
+        return result;
+      } catch (error) {
+        throw new NotFoundException(`Role with ID ${id} not found`);
+      }
```
medtaher123, line 175: You don’t need both checks. The second check already looks for all users, including the active one, so the first check is unnecessary. You can simplify by just using the second check to cover everything.

medtaher123, line 175:
```suggestion
        const result = await this.roleService.deleteOne(id);
        if (result.deletedCount === 0) {
          throw new NotFoundException(`Role with ID ${id} not found`);
        }
        return result;
```
The throw new `NotFoundException` is redundant, and the `try-catch` block isn’t necessary here. If a different error occurs, returning a `NotFoundException` could be misleading, as we should return an `InternalServerErrorException` or another appropriate error instead.

medtaher123 (changes requested):

api/src/user/controllers/role.controller.ts
medtaher123: Actually he doesn't need to know that he is the user using the role. The only information he needs is that some user is still using the role. 
This is still a suggestion for the behaviour, we have to check with @marrouchi  if this is the right think to do

marrouchi (approved):

PR #67, opened 2024-09-22
Add permanent option context var 

### LINKED ISSUE #48: 💡 [REQUEST] - Context vars : Permanent option
### Start Date

_No response_

### Implementation PR

_No response_

### Reference Issues

_No response_

### Summary

This is an important feature to implement, we need to add a flag for context variables called "permanent" boolean. Currently a context var value is only kept during the current conversation and then gets lost. It would be good to add the "permanent" option allow the value to be stored in the subscriber object in a attribute called "context" (json).

### Basic Example

- Add "permanent" attribute to the context var entity (boolean)
- Add "context" attribute in the subscriber entity (JSON object)

### Drawbacks

We cannot typecheck the user context object.

### Unresolved questions

_No response_

medtaher123:
# Motivation

We need to add a flag for context variables called "permanent" boolean. Currently a context var value is only kept during the current conversation and then gets lost. It would be good to add the "permanent" option allow the value to be stored in the subscriber object in a attribute called "context" (json).

## Feature desciption:
There are two types of context variables in Hexabot: non-permanent and permanent.

### 1. Non-Permanent Context Variables

- **Temporary Storage:** Non-permanent context variables are stored in the conversation context.
- **Scope:** They are specific to the ongoing conversation and are lost once the conversation ends. This makes them ideal for storing temporary data that is relevant only during the current interaction.
- **Example:** If a bot asks a user for their preferred meeting time during a session, the bot may store the response as a non-permanent context variable to use in the rest of the conversation.

### 2. Permanent Context Variables

- **Persistent Storage:** Permanent context variables are stored in the subscriber context, which means they persist across multiple conversations.
- **Scope:** These variables retain their values even after a conversation has ended, allowing the bot to remember information about a user across different interactions.
- **Example:** If a user provides a preference for a certain feature, such as opting into notifications, the bot can store this preference permanently. The bot will retain this information for future use, without needing to ask the user again in subsequent conversations.

Fixes #48 

## Type of change

# Checklist:


api/src/chat/services/conversation.service.ts
```diff
@@ -182,6 +206,13 @@ export class ConversationService extends BaseService<
           'Conversation Model : No conversation has been updated',
         );
       }
+
+      //TODO: add check if nothing changed don't update
+
+      this.subscriberService.updateOne(convo.sender, {
```
marrouchi, line 212:
```suggestion
      await this.subscriberService.updateOne(convo.sender, {
```

api/src/chat/services/conversation.service.ts
```diff
@@ -79,13 +84,26 @@ export class ConversationService extends BaseService<
...
+      await this.contextVarService.find({
+        name: { $in: next.capture_vars.map((c) => c.context_var) },
+      })
+    ).reduce(
+      (acc, cv) => {
+        acc[cv.name] = cv;
+        return acc;
+      },
+      {} as Record<string, ContextVar>,
+    );
```
marrouchi, line 105: I suggest we add this method to the "contextVarService" : getContextVarsByBlock(next)

marrouchi (changes requested):
Looks good @medtaher123 Thanks 👏  

I think you will need to adapt the unit tests.

api/src/chat/services/conversation.service.ts
medtaher123: Yes, I'm aware of that. Thank you for pointing it out

medtaher123: Thank you for the suggestion. However, I think we don't need to await here, as making the function async allows the update to happen in the background. This way, we can continue processing other tasks without waiting for it to complete, which can improve performance, especially if the update isn't critical for the next steps

medtaher123:
- [ ] Add "permanent" attribute to the context var entity 
- [ ] Add "context" attribute in the subscriber entity
- [ ] Capture permanent context vars in the subscriber
- [ ] correct existing unit tests
- [ ] add new corresponding unit tests
- [ ] Use the permanent context vars in the token replacement
- [ ] add unit tests
- [ ] add the permanent option in the frontentend contextVars page

api/src/chat/services/conversation.service.ts
marrouchi: We need a await I think cause the next logic may need the subscriber to be up-to-date.

api/src/chat/dto/context-var.dto.ts
```diff
@@ -20,6 +20,9 @@ export class ContextVarCreateDto {
   @IsNotEmpty()
   @IsString()
   name: string;
+
+  @ApiProperty({ description: 'Is context var permanent', type: Boolean })
```
marrouchi, line 24:
```suggestion
  @IsOptional()
  @IsBoolean()
  @ApiProperty({ description: 'Is context var permanent', type: Boolean })
```

marrouchi (changes requested):

marrouchi:
@medtaher123 Thank you 🙏  Can you please add a comment to the field permanent to explain what it does ?
Also, in the PR, add a detailed description of what a permanent context var is compared to a non permanent. Will be used in the documentation ;)

marrouchi (approved):

PR #81, opened 2024-09-24
fix: hiding of other manager roles in case of deletion

devamitranjan:
# Motivation

Fixed hiding of other manager roles in case of deletion of a single role.

Fixes # (issue)

## Type of change

# Checklist:


devamitranjan:
[Screencast from 24-09-24 05:17:28 PM IST.webm](https://github.com/user-attachments/assets/007b35b3-d8aa-41a2-818e-88d2743f88d8)
Added the recording for the reference.

Emnaghz (approved):
Tested successfully, and it behaves as expected. :+1:

devamitranjan:
@marrouchi Can you please review this too?

frontend/src/hooks/crud/useDelete.tsx
```diff
@@ -58,11 +58,12 @@ export const useDelete = <
...
         queryClient.removeQueries({
           predicate: ({ queryKey }) => {
-            const [qType, qEntity] = queryKey;
+            const [qType, qEntity, qId] = queryKey;
 
             return (
               (qType === QueryType.count || qType === QueryType.collection) &&
-              isSameEntity(qEntity, entity)
+              isSameEntity(qEntity, entity) &&
+              qId === id
```
marrouchi, line 66: Thank you @devamitranjan 

There's 3 types of state objects : 
- count : holds the count for a given data grid
- collection : usually used to store the ids of a given page of a data grid
- item : store individual objects.

useDelete is used to delete a single item, so in this case the id would be the id of the object deleted.

Maybe I'm getting something wrong, could you please explain further (an example would be good).

marrouchi (changes requested):

frontend/src/hooks/crud/useDelete.tsx
marrouchi: This is should be the fix : https://github.com/Hexastack/Hexabot/pull/85

Let me know if it does make sense.

devamitranjan: You're absolutely right, @marrouchi . Adding this condition was preventing the queries from becoming stale and triggering the API call to fetch fresh data. As a result, the count was not being updated correctly.

Instead of using removeQueries, we should use invalidateQueries so that the data remains cached until the fresh data is fetched from the API in the background. This way, other existing data won't be removed from the cache, preventing unintended data loss.

Added the change.

Issue #127, opened 2024-10-02
🐛 [BUG] - Update of the content in the cms is getting failed


devamitranjan:
### Description

When updating through toggle button or content dialog the status is not getting updated.
[problem.webm](https://github.com/user-attachments/assets/bd166f79-0f19-4713-8a0d-6f21cd240f5e)

### Reproduction steps

```bash
1. Go to 'Manage Content/CMS/Content Type'
2. Click on 'toggle button for the status in the listing page or inside modal'
3. Change the state.
4. See error
```

devamitranjan:
@marrouchi Can you please assign this issue to me? I don't have access to assign it to myself.
Also, can you please add the hacktoberfest label if possible.

marrouchi:
Hello @devamitranjan done !

PR #128, opened 2024-10-02
Content update

devamitranjan:
# Motivation
Fixed the failing of the status in the content in the cms.
Fixes https://github.com/Hexastack/Hexabot/issues/127

## Type of change

# Checklist:


marrouchi (changes requested):
Hello @devamitranjan 

Thank you, I think we need to fix this differently. 

Currently the  controller path operation is using the ContentCreateDto

```
@CsrfCheck(true)
  @Patch('/:id')
  async updateOne(
    @Body() contentDto: **ContentCreateDto**,
    @Param('id') id: string,
  ): Promise<Content> {
```

We need to add another Dto for updates when we have two use cases : 
- Either it's to update everyting : title, status, fields
- Just update the status attribute

PR #271, opened 2024-10-26
fix(frontend): visual editor pattern UI representation

### LINKED ISSUE #270: 🐛 [BUG] - blockForm dialog regex pattern extra slashes bug
### Description

We can notice the bug by watching the video bellow or by following the reproducing steps  
<video src='https://github.com/user-attachments/assets/7623e188-e804-4db0-940d-c1b58fa36e96' />

### Reproduction steps

```bash
1. Go to Visual editor
2. Create a block
3. Editor the block to add a regex pattern
4. Save the block
5. Click again to edit the block
6. Close the dialog component
7. See that the pattern is changing in the UI
```

### Logs

_No response_

### Browsers

Chrome

### OS

Linux

yassinedorbozgithub:
# Motivation
The main motivation of this PR is to fix a bug existing actually in the Visual Editor. 
The block pattern field is displayed incorrectly in the UI.
More details are attached to the issue related to this PR.

Fixes #270

# Screenshots:
Cases fully support by this proposed change :
- Invalid regex case
image
- Empty regex case
image
- Edge case pattern that can bypass the check logic
image

# Type of change:

# Checklist:

### DIFF
--- a/frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
+++ b/frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
@@ -32,7 +32,12 @@ import { ContentPostbackInput } from "./ContentPostbackInput";
 import { PostbackInput } from "./PostbackInput";
 
 const isRegex = (str: Pattern) => {
-  return typeof str === "string" && str.startsWith("/") && str.endsWith("/");
+  return (
+    typeof str === "string" &&
+    str.startsWith("/") &&
+    str.endsWith("/") &&
+    str.length > 1
+  );
 };
 const getType = (pattern: Pattern): PatternType => {
   if (isRegex(pattern)) {
--- a/frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
+++ b/frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
@@ -268,19 +268,14 @@ const PatternInput: FC<PatternInputProps> = ({
             {...registerInput(t("message.regex_is_invalid"), idx, {
               validate: (pattern) => {
                 try {
-                  if (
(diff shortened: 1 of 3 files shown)


frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
```diff
@@ -268,19 +268,14 @@ const PatternInput: FC<PatternInputProps> = ({
             {...registerInput(t("message.regex_is_invalid"), idx, {
               validate: (pattern) => {
                 try {
-                  if (
-                    pattern.at(0) === "/" &&
-                    pattern.at(-1) === "/" &&
-                    typeof pattern === "string"
-                  )
-                    new RegExp(pattern.slice(1, -1));
+                  if (isRegex(pattern)) new RegExp(pattern.slice(1, -1));
```
marrouchi, line 271: If it doesn't starts/ends with "/", this will bypass the check. 

I think it should be 
```suggestion
                  new RegExp(isRegex(pattern) ? pattern.slice(1, -1) : pattern);
```

marrouchi (changes requested):

frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
yassinedorbozgithub: Hi @marrouchi, i think that we use react-hook-form setValueAs register option that will executed before the validation process to be sure that the pattern is always starting and ending with '/' if i'm not mistaking.
image
 Reference:
 https://react-hook-form.com/docs/useform/registe

IkbelTalebHssan (approved):

marrouchi (approved):

PR #287, opened 2024-10-28
Train use mlflow

MohamedAliBouhaouala:
# Motivation

Fixes # (issue)

# Type of change:

# Checklist:


nlu/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: '3.9'
```
marrouchi, line 1: Please add these ROOT/docker/docker-compose.nlu.yml and ROOT/docker/docker-compose.nlu.dev.yml

nlu/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      dockerfile: Dockerfile  # Specify the Dockerfile explicitly
+    image: mlflow
+    container_name: mlflow_server
+    environment:
+      - BACKEND_STORE_URI=postgresql://postgres:postgres@mlflow_postgres:5432/mlflow_db  # Connection string to Postgres
+      - ARTIFACT_STORE_URI=./mlruns  # Local directory for storing artifacts
+    ports:
+      - "5002:5000"  # Expose MLflow UI
+    volumes:
+      - ./mlruns:/mlruns  # Mount local directory for MLflow artifacts
```
marrouchi, line 28:
```suggestion
      - mlruns:/mlruns  # Mount local directory for MLflow artifacts
```
Add "mlruns:" to the volumes

nlu/models/intent_classifier.py
```diff
@@ -34,6 +36,7 @@
     'fr': "dbmdz/bert-base-french-europeana-cased",
 }
 
+mlflow.set_tracking_uri("http://0.0.0.0:5002")
```
marrouchi, line 39:
```suggestion
mlflow.set_tracking_uri("http://0.0.0.0:5002")
```
This should be the docker hostname

nlu/models/intent_classifier.py
```diff
@@ -119,35 +122,70 @@ def format_scores(self, scores: Dict[str, dict]):
...
-        super().fit(
-            x, encoded_intents, epochs=self.hparams.num_epochs, batch_size=32, shuffle=True)
-
-        # Persist the model
-        self.extra_params["intent_names"] = intent_names
-
-        self.save()
-
+        # Start MLflow run
+        with mlflow.start_run() as run:
```
marrouchi, line 126: Shouldn't this be implement in the parent class so that it would work for all models ?

nlu/run.py
```diff
@@ -96,7 +96,7 @@
     if os.path.isfile(os.path.join(model.save_dir, "checkpoint")):
         model.restore()
     else:
-        model.save()
+        model.save_model()
```
marrouchi, line 99: Not sure why we need to rename this.

MohamedAliBouhaouala: To avoid confusion with keras' built-in save method. It's been throwing aberrant exceptions. Solved it by renaming the boilerplate method

nlu/models/intent_classifier.py
MohamedAliBouhaouala: Ideally it should be implemented as a helper class. Otherwise, we're logging metrics during training. The base class doesn't have a fit method.

PR #356, opened 2024-11-21
Add HuggingFace login functionality

medchedli:
Implement login functionality for HuggingFace using an access token from environment variables.


marrouchi (approved):
Awesome @medchedli thanks 👏 !

Issue #109, opened 2024-09-30
🐛 [BUG] - API Translation Refresh - Custom blocks settings


marrouchi:
### Description

Currently the API endpoint for refreshing translations extract all custom blocks settings (including secret). We need to restrict the translation refresh to only text like settings.

### Reproduction steps

```bash
1. Login and go to Visual Editor
2. Add the RAG custom block
3. Edit the block and go to "Message"
4. Update token
5. Go to "Localization / Translations"
6. Hit Refresh
7. You would be able to see the secret.
```

kartik1112:
Hey @marrouchi can you assign this to me?

marrouchi:
@kartik1112 Done 👍

kartik1112:
![image](https://github.com/user-attachments/assets/e9926881-ddb5-4f4e-9efb-b58ffa91332a)

@marrouchi I can't see RAG custom block here.. Can you help?
and if i go for dummy I see empty space in messages tab

marrouchi:
Hello @kartik1112, please do git pull latest from main and then install it : 
`npm install`
`npx hexabot install plugin Hexastack/hexabot-plugin-chatgpt`
It has been moved in a seperate repo and became installable. Sorry for the inconvinence.

marrouchi:
@AuraOfDivinity Task has been assign to you since we didn't get any response from @kartik1112

PR #360, opened 2024-11-22
fix: display translatable settings only in translations UI #109

abdou6666:
# Motivation

The issue was that some plugin settings, which are not translatable, were being returned from the backend alongside translatable ones. The fix ensures that only the translatable settings are included in the backend response for the translations UI, preventing non-translatable settings from being unnecessarily returned and displayed.

Fixes # (issue)

# Type of change:

# Checklist:

### DIFF
--- a/api/src/i18n/services/translation.service.spec.ts
+++ b/api/src/i18n/services/translation.service.spec.ts
@@ -73,7 +73,7 @@ describe('TranslationService', () => {
             find: jest.fn().mockResolvedValue([
               {
                 translatable: false,
-                global_fallback: 'global_fallback',
+                label: 'global_fallback',
                 fallback_message: ['Global fallback message'],
               },
             ]),
@@ -175,7 +175,6 @@ describe('TranslationService', () => {
         fallback_message: ['Global fallback message'],
       },
     } as Settings);
-
     const strings = await service.getSettingStrings();
     expect(strings).toEqual([]);
   });
--- a/api/src/i18n/services/translation.service.spec.ts
+++ b/api/src/i18n/services/translation.service.spec.ts
@@ -22,7 +22,6 @@ import { TranslationService } from '../services/translation.service';
 
 describe('TranslationService', () => {
   let service: TranslationService;
-  let settingService: SettingService;
   let i18nService: I18nService;
   let pluginService: PluginService;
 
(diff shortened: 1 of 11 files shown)


api/src/i18n/services/translation.service.spec.ts
```diff
@@ -58,12 +67,16 @@ describe('TranslationService', () => {
...
+              .fn()
+              .mockResolvedValue(['Global fallback message']),
+            find: jest.fn().mockResolvedValue([
+              {
+                translatable: true,
+                key: 'global_fallback',
+                value: 'Global fallback message',
               },
-            }),
+            ]),
```
marrouchi, line 79: Is this needed ? 

+ in settings we don't have "key"

api/src/plugins/types.ts
```diff
@@ -24,7 +24,9 @@ export interface CustomBlocks {}
 
 type BlockAttrs = Partial<BlockCreateDto> & { name: string };
 
-export type PluginSetting = Omit<SettingCreateDto, 'weight'>;
+export type PluginSetting = Omit<SettingCreateDto, 'weight'> & {
+  translatable?: boolean;
+};
```
marrouchi, line 29: Since it's already in DTO, we don't need to redefine it here.

api/src/utils/test/fixtures/setting.ts
```diff
@@ -19,69 +19,87 @@ export const settingFixtures: SettingCreateDto[] = [
...
+    weight: 10,
+    translatable: true,
   },
   {
     group: 'contact',
     label: 'company_name',
     value: 'Your company name',
     type: SettingType.text,
     weight: 2,
+    translatable: false,
```
marrouchi, line 38: Let's just not add it, if it doesn't exist, it would mean it's false.

marrouchi (changes requested):
Awesome @abdou6666 , left some comments 👏

api/src/i18n/services/translation.service.spec.ts
abdou6666: I think it's needed since it's being used in the refresh endpoint. When I remove it, the test 'should return an empty array from the settings when global fallback is disabled' doesn't pass

api/src/i18n/services/translation.service.spec.ts
```diff
@@ -58,12 +67,16 @@ describe('TranslationService', () => {
...
-            getSettings: jest.fn().mockResolvedValue({
-              chatbot_settings: {
-                global_fallback: true,
+            getSettings: jest
+              .fn()
+              .mockResolvedValue(['Global fallback message']),
+            find: jest.fn().mockResolvedValue([
+              {
+                translatable: false,
+                global_fallback: 'global_fallback',
```
marrouchi, line 76:
```suggestion
                label: 'global_fallback',
```
Focus is the bridge between intention and accomplishment 🙏

api/src/utils/test/fixtures/setting.ts
```diff
@@ -20,6 +20,14 @@ export const settingFixtures: SettingCreateDto[] = [
...
     weight: 1,
   },
+  {
+    group: 'contact',
+    label: 'greeting',
+    value: 'hello',
+    type: SettingType.text,
+    weight: 10,
+    translatable: true,
+  },
```
marrouchi, line 30: Do we need this for a given unit test ?

If yes, why don't we use existing ones ?

marrouchi (changes requested):

api/src/i18n/services/translation.service.spec.ts
abdou6666: Done

api/src/utils/test/fixtures/setting.ts
abdou6666: those are unnecessary i removed them

medchedli:
#363

PR #444, opened 2024-12-13
Fix : Number settings gets saved as strings in MongoDB

### LINKED ISSUE #427: 🤔 [ISSUE] - Settings of type number get stored as string
### Description

Settings of type number get stored as strings. We may have the same issue for boolean.

### Reproduction steps

```bash
-
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

abdou6666:
# Motivation

This PR fixes the issue when updating a setting field that should be saved as Number and saved as String in MongoDB because mongoose middleware/hooks does not run by default on findOneAndUpdate

Fixes #427

# Type of change:

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/api/src/setting/repositories/setting.repository.ts
+++ b/api/src/setting/repositories/setting.repository.ts
@@ -12,10 +12,19 @@ import {
   IHookSettingsGroupLabelOperationMap,
 } from '@nestjs/event-emitter';
 import { InjectModel } from '@nestjs/mongoose';
-import { Document, Model, Query, Types } from 'mongoose';
+import {
+  Document,
+  FilterQuery,
+  Model,
+  Query,
+  Types,
+  UpdateQuery,
+  UpdateWithAggregationPipeline,
+} from 'mongoose';
 
 import { I18nService } from '@/i18n/services/i18n.service';
-import { BaseRepository } from '@/utils/generics/base-repository';
+import { BaseRepository, EHook } from '@/utils/generics/base-repository';
+import { TFilterQuery } from '@/utils/types/filter.types';
 
 import { Setting } from '../schemas/setting.schema';
 import { SettingType } from '../schemas/types';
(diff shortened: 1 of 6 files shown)


frontend/src/components/settings/index.tsx
```diff
@@ -133,8 +133,14 @@ export const Settings = () => {
...
     const subscription = watch((values, { name }) => {
       const [group, label] = (name as string).split(".");
-
-      debouncedUpdate(group, label, values[group][label]);
+      const currentSettingValue = values[group][label] as any;
+      // if setting value is Number but typed string convert it to number
+      const settingValue =
+        typeof values[group][label] === "string" && isNaN(currentSettingValue)
+          ? values[group][label]
+          : Number(values[group][label]);
```
yassinedorbozgithub, line 141:
```suggestion
       const setting = settings.find(
        (setting) => setting.group === group && setting.label === label,
      )?.value;
      // if setting value is Number but typed string convert it to number
      const settingValue =
        typeof setting === "number"
          ? Number(values[group][label])
          : values[group][label];
```
Hi @abdou6666, do you think maybe that we can do something like this to resolve the data save issue

frontend/src/components/settings/index.tsx
```diff
@@ -133,8 +133,14 @@ export const Settings = () => {
   useEffect(() => {
     const subscription = watch((values, { name }) => {
       const [group, label] = (name as string).split(".");
-
-      debouncedUpdate(group, label, values[group][label]);
+      const currentSettingValue = values[group][label] as any;
```
marrouchi, line 136: Let's rather use the rules in the "Controller" : 
```
rules={{ valueAsNumber: setting.type === 'number'}}
```

marrouchi (changes requested):

api/src/setting/repositories/setting.repository.ts
```diff
@@ -100,4 +99,62 @@ export class SettingRepository extends BaseRepository<Setting> {
...
+    } else if (type === SettingType.attachment) {
+      if (typeof value !== 'string' && typeof value !== null) {
+        throw new Error(
+          'Setting Model (attachement): Value must be a string or null !',
+        );
+      }
+    } else if (type === SettingType.secret && typeof value !== 'string') {
+      throw new Error('Setting Model (secret) : Value must be a string');
+    } else if (type === SettingType.select && typeof value !== 'string') {
+      throw new Error('Setting Model (select): Value must be a string array!');
```
yassinedorbozgithub, line 153:
```suggestion
      throw new Error('Setting Model (select): Value must be a string');
```

yassinedorbozgithub (changes requested):

PR #482, opened 2024-12-20
feat: change scope

medtaher123:
This PR changes the `LoggerService` scope from `singleton` to `transient`. The main motivation for this change is to allow adding context-specific data for each class that uses the `LoggerService`.By making the service `transient`, a new instance of `LoggerService` will be created for each injection, enabling us to set a unique context for logging per class.**Implications**  
- **Pros** :
  - Enables class-specific logging contexts, improving traceability and debugging.

  - More flexible for future logging enhancements.
 
- **Cons** :
  - Loss of singleton behavior means there will be multiple instances, potentially increasing memory usage slightly.

### DIFF
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -84,9 +84,9 @@ async function bootstrap() {
     app.useWebSocketAdapter(redisIoAdapter);
   }
 
-  process.on('uncaughtException', (error) => {
+  process.on('uncaughtException', async (error) => {
     if (error.stack.toLowerCase().includes('smtp'))
-      app.get(LoggerService).error('SMTP error', error.stack);
+      (await app.resolve(LoggerService)).error('SMTP error', error.stack);
     else throw error;
   });
 
@@ -95,7 +95,7 @@ async function bootstrap() {
     swagger(app);
   }
 
-  await app.listen(3000);
+  await app.listen(4001);
 }
 
 bootstrap();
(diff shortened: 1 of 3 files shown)


api/src/main.ts
```diff
@@ -86,7 +86,7 @@ async function bootstrap() {
 
   process.on('uncaughtException', (error) => {
```
MohamedAliBouhaouala, line 87: process.on('uncaughtException', async (error) => {
    if (error.stack?.toLowerCase().includes('smtp'))
      (await (app.resolve(LoggerService))).error('SMTP error', error.stack);
    else throw error;
  });

api/src/main.ts
```diff
@@ -86,7 +86,7 @@ async function bootstrap() {
```
MohamedAliBouhaouala, line 86: process.on('uncaughtException', async (error) => {
    if (error.stack?.toLowerCase().includes('smtp'))
      (await (app.resolve(LoggerService))).error('SMTP error', error.stack);
    else throw error;
  });

api/src/main.ts
```diff
@@ -86,7 +86,7 @@ async function bootstrap() {
 
   process.on('uncaughtException', (error) => {
     if (error.stack?.toLowerCase().includes('smtp'))
-      app.get(LoggerService).error('SMTP error', error.stack);
+      (await app.resolve(LoggerService)).error('SMTP error', error.stack);
```
MohamedAliBouhaouala, line 89: process.on('uncaughtException', async (error) => {
    if (error.stack?.toLowerCase().includes('smtp'))
      (await (app.resolve(LoggerService))).error('SMTP error', error.stack);
    else throw error;
  });

api/src/main.ts
abdou6666, line 87: the callback function must be async

abdou6666 (changes requested):

PR #549, opened 2025-01-09
fix(api): optimize performance in preUpdateMany

### LINKED ISSUE #536: 🤔 [ISSUE] - (N +1 query problem) preUpdateMany block.repository.ts
### Description

current implementation of preUpdateMany loops through the ids and uses findOne on each of the ids , we can use this this.repository.find({ _id: { $in: ids } }); instead to decrease number of queries & improve performance , we might need to adapt the logic & write new tests

omarNaifer12:
# Motivation

Updated the function prepareBlocksInCategoryUpdateScope to replace multiple findOne API calls with a single API call. Updated its tests accordingly

Fixes #536

# Type of change:

# Checklist:


abdou6666 (approved):

MohamedAliBouhaouala (approved):

api/src/chat/repositories/block.repository.ts
```diff
@@ -177,18 +177,21 @@ export class BlockRepository extends BaseRepository<
     category: string,
     ids: string[],
   ): Promise<void> {
-    for (const id of ids) {
-      const oldState = await this.findOne(id);
-      if (oldState && oldState.category !== category) {
-        const updatedNextBlocks = oldState.nextBlocks?.filter((nextBlock) =>
+
+    const oldStates: Block[] = await this.find({ _id: { $in: ids } });
```
yassinedorbozgithub, line 181:
```suggestion
    const oldStates = await this.find({ _id: { $in: ids } });
```
The find method will return the expected type **Block[]**

yassinedorbozgithub (changes requested):

api/src/chat/repositories/block.repository.ts
marrouchi, line 181: @omarNaifer12 thanks for your contribution.

In addition to @yassinedorbozgithub comment, I suggest adding the criteria "different category" so that we avoid the "if" condition inside the loop.

marrouchi (changes requested):

api/src/chat/repositories/block.repository.ts
omarNaifer12: Thank you @marrouchi for the feedback.

yassinedorbozgithub:
Closed in favor of #919

PR #552, opened 2025-01-11
fix(api): execute fallback message when no Global Fallback

### LINKED ISSUE #115: 🤔 [ISSUE] -  Block-Level Fallbacks Are Ignored When Global Fallback is Disabled
### Description

 When the global fallback option is disabled in the settings, the chatbot does not respect the fallback configurations set at the block level. This leads to the chatbot failing to respond appropriately when no matching intent is found within a block.

### Reproduction steps

```bash
1. Go to 'Setings > Chatbot'
2. Disable Global Fallbacks
3. Go back to Visual Editor, create an entry-point block and add a new fallback in 'Options' section
4. Start chatting with the chatbot and try to enter a message out of the context
5. No fallback message is displayed
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

omarNaifer12:
# Motivation
I just added an extra condition to check if fallback_block is not nullable, to ensure it dont enter to the  condition and trigger catch block if it's nullable.If you don't want to duplicate the logic of sendMessageToSubscriber, I can remove it from the catch block.

Fixes #115 

# Type of change:

# Checklist:

### DIFF
--- a/api/src/chat/services/bot.service.ts
+++ b/api/src/chat/services/bot.service.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
  *
  * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
  * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
@@ -454,7 +454,8 @@ export class BotService {
           this.logger.debug('No message blocks available!');
           if (
             settings.chatbot_settings &&
-            settings.chatbot_settings.global_fallback
+            settings.chatbot_settings.global_fallback &&
+            settings.chatbot_settings.fallback_block
           ) {
             this.eventEmitter.emit('hook:analytics:fallback-global', event);
             this.logger.debug('Sending global fallback message ...');
@@ -485,7 +486,22 @@ export class BotService {
               } as any as BlockFull);
             }
           }
-          // Do nothing ...
+          this.sendMessageToSubscriber(event, {
+            id: 'global-fallback',
+            name: 'Global Fallback',
+            message: settings.chatbot_settings.fallback_message,
+            options: {},
+            patterns: [],
+            assign_labels: [],
+            starts_conversation: false,
+            position: { x: 0, y: 0 },
+            capture_vars: [],
+            builtin: true,
+            createdAt: new Date(),
+            updatedAt: new Date(),
+            attachedBlock: null,
+          } as any as BlockFull);
+
           return;
         }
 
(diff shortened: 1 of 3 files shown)


IkbelTalebHssan (changes requested):
Hi @omarNaifer12 
Thank you for contributing to Hexabot.  We really appreciate the effort you’ve put into this.
As seen in the screenshots below after disabling the global fallback the block- level one is not displayed.
Screenshot from 2025-01-20 14-56-50
Screenshot from 2025-01-20 14-55-19

abdou6666 (approved):

api/src/chat/services/bot.service.ts
```diff
@@ -485,7 +486,22 @@ export class BotService {
...
+            assign_labels: [],
+            starts_conversation: false,
+            position: { x: 0, y: 0 },
+            capture_vars: [],
+            builtin: true,
+            createdAt: new Date(),
+            updatedAt: new Date(),
+            attachedBlock: null,
+          } as any as BlockFull);
+
```
abdou6666, line 504: Great job on this implementation! I think it would be even better if you could refactor the code to avoid duplication in the `catch` block. By using a `finally` statement to send the message to the subscriber instead

abdou6666 (changes requested):

marrouchi:
I think there's a misunderstanding about the issue, let's pause here until we validate the issue.

omarNaifer12:
Hi @IkbelTalebHssan, thank you for the feedback! Could you please clarify the purpose of enabling the fallback option in the box and what its message should achieve?

omarNaifer12:
Here is my understanding of issue #115:
When we send a message to the chatbot and it doesn't match any box it first checks the global fallback in the settings to return the fallback block. If the global fallback is not enabled the chatbot doesn't send anything.
I have resolved the issue based on this condition only. Please let me know if my understanding of the issue is correct or if there are additional problems I may have missed

api/src/chat/services/bot.service.ts
omarNaifer12: Thanks! I'm not sure where to include the finally statement but I noticed that if we remove the sendMessageToSubscriber from the catch block it works because if there is an error, the catch block doesn't return anything to close the function.

IkbelTalebHssan:
> Hi @IkbelTalebHssan, thank you for the feedback! Could you please clarify the purpose of enabling the fallback option in the box and what its message should achieve?

Hi @omarNaifer12, 
The purpose of the block-level fallback message is to provide a customized response specific to a particular block when the user's input does not meet the expected criteria. This fallback is triggered when the user sends a message that cannot be accepted as valid input for that specific block.

For example, imagine you have configured a block that requires the user to input only numbers for a specific purpose, such as entering a phone number, age, or weight. If the user attempts to trigger this block by sending a message containing letters, special characters, or any invalid input, the block-level fallback message will be activated. This ensures that the user receives a tailored response explaining the issue, rather than triggering the global fallback (if one exists) or receiving no response at all (in cases where the global fallback is disabled).

In essence, the block-level fallback message serves as a context-aware error handler, improving user experience by providing clear, relevant feedback when their input does not align with the block's requirements.

omarNaifer12:
Thank you @IkbelTalebHssan,I will handle this case as well.

yassinedorbozgithub:
Closed in favor of https://github.com/Hexastack/Hexabot/pull/944

PR #560, opened 2025-01-13
fix: handover & handback should update assignedAt attribute

### LINKED ISSUE #559: 🐛 [BUG] - handover & handback do not update assignedAt attribute
### Description

-

### Reproduction steps

```bash
-
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

abdou6666:
# Motivation

When handing over happens we should update the assignedAt field in DB to a new Date, and when handback happens we should reset it to null 

Fixes #559

# Type of change:

# Checklist:

### DIFF
--- a/api/src/chat/repositories/subscriber.repository.ts
+++ b/api/src/chat/repositories/subscriber.repository.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
  *
  * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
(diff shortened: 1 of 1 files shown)


MohamedAliBouhaouala (commented):
Can you provide an example of why is this necessary? I'm kinda out of context

abdou6666:
> Can you provide an example of why is this necessary? I'm kinda out of context

when handover happens we've to update assignedTo to 'userId' and assignedAt to a new date
when handback happens we need to reset them (maybe keep assignedAt i'm still not sure about that)

MohamedAliBouhaouala (approved):

abdou6666:
closed in favor of #566

Issue #354, opened 2024-11-21
🐛 [BUG] - Move block between categories


abdou6666:
### Description

Issue happens when user selects block and links and the last item selected is a block the 'move' button becomes enabled 

### Reproduction steps

```bash
1. select block
2. select link
3. select block

Move button becomes enabled again
when you click on move button api returns 500
```

RumaniSadhukhan170204:
@medchedli please assign me to this

PR #571, opened 2025-01-15
Fixed Issue: 'Move' Button Becomes Enabled After Selecting Block and Link, Leading to 500 API Error (#354)

RumaniSadhukhan170204:
This bug occurred when a user selected a block, then a link, and then selected a block again. In this sequence, the "Move" button became enabled, despite the selection not being valid for moving. When the user clicked the "Move" button, it triggered a 500 internal server error in the API.

Steps to Reproduce:

Select a block.
Select a link.
Select a block again.
Expected Behavior:
The "Move" button should remain disabled when an invalid selection is made.

Actual Behavior:
The "Move" button becomes enabled, and clicking it triggers a 500 API error.

Fix:
The issue was resolved by ensuring that the "Move" button only becomes enabled for valid selections, preventing the API error from being triggered.

### DIFF
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -1,0 +1,49 @@
+import express, { Request, Response } from 'express';
+import { Block } from './models/Block';
+import { Category } from './models/Category';
+
+const app = express();
+const port = 3000;
+
+
+app.use(express.json());
+
+
+app.post('/move-block', async (req: Request, res: Response) => {
+  const { block_id, category_id } = req.body;
+
+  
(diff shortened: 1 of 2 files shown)


marrouchi:
Hello @RumaniSadhukhan170204 , thank you for the attempt to contribute but we cannot merge it cause the API uses Nestjs and so any contribution needs to abide by the tech stack we are using. I suggest that begin with small issues (tagged with "good first issue"). See you around!

PR #645, opened 2025-01-29
fix: logout disconnection sockets

### LINKED ISSUE #644: 🐛 [BUG] - logout API sockets disconnection
### Description

# Description
We can notice that if we open many tabs and disconnect from one the others Web-sockets will stay connected 

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
Currently if we logout after opening many only one tab will disconnect. This PR main motivation is to resolve the mentioned bug 

Fixes #644

# Fix demonstration + How to test
<video src="https://github.com/user-attachments/assets/db51d46b-a755-4230-a17e-47ebae88fa45"></video>

# Type of change:
# Checklist:

### DIFF
--- a/api/src/user/controllers/auth.controller.ts
+++ b/api/src/user/controllers/auth.controller.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
  *
  * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
  * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
@@ -11,6 +11,8 @@ import {
   Body,
   Controller,
   Get,
+  Headers,
+  Inject,
   InternalServerErrorException,
   Param,
   Post,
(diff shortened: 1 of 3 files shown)


yassinedorbozgithub:
> I've added a log statement to check for number of socket connections & i got this after I logged out image

This means that the logic works as expected, all sockets are closed

abdou6666:
but on connection we have weird behaviour. 
we getting x2 number of socket & we aren't running in strict mode

image

abdou6666:
> > I've added a log statement to check for number of socket connections & i got this after I logged out image
> 
> This means that the logic works as expected, all sockets are closed

yes!

abdou6666:
> but on connection we have weird behaviour. we getting x2 number of socket & we aren't running in strict mode
> 
> image

after abit of testing this might be happening due to nextjs running the code server side first.

i added this code to `socket-hooks.tsx` in the socket provider useEffect 
`
if (typeof window !== "undefined") {
      // Ensure client-side execution
      return;
    }
` 
and i got this instead we should further investigate 
image

yassinedorbozgithub:
> socket-hooks.tsx

Nice catch, if it's directly related to the API logout we can included in the PR 
Otherwise, i recommend to create a dedicated issue for it

abdou6666:
> > but on connection we have weird behaviour. we getting x2 number of socket & we aren't running in strict mode
> > image
> 
> after abit of testing this might be happening due to nextjs running the code server side first.
> 
> i added this code to `socket-hooks.tsx` in the socket provider useEffect `if (typeof window !== "undefined") { // Ensure client-side execution return; }` and i got this instead we should further investigate image
> 
> Nice catch, if it's directly related to the API logout we can included in the PR Otherwise, i recommend to create a dedicated issue for it

agreed !

PR #646, opened 2025-01-29
feat: Cleanup orphan extensions' settings

yassine-sallemi:
# Motivation

Unnecessary settings of orphan extensions are now deleted to avoid errors in frontend

# Type of change:

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas


marrouchi (changes requested):
I suggest we make the service part of the setting module, let's call it CleanUpService.

Let's for now handle only channels and helpers.

api/src/extra/extension-cleanup.service.ts
```diff
@@ -0,0 +1,73 @@
...
+      debugger;
+      if (extensionType in ExtensionType && !activeExtensions.includes(group)) {
+        this.logger.debug(`Deleting orphaned settings for ${group}...`);
+        await this.settingService.deleteGroup(group);
+      }
+    }
+  }
+
+  private getHelperNames(): string[] {
+    return this.helperService.getAll().map((handler) => handler.getName());
```
marrouchi, line 63:
```suggestion
    return this.helperService.getAll().map((handler) => handler.getNamespace());
```
Use getNamespace() to avoid replacing "-" by "_".

marrouchi (changes requested):

yassine-sallemi:
I think that making the new `CleanupService` part of the `SettingModule` will cause a problem, as it is loaded before the extensions are loaded, so it may delete some saved settings of some extensions, and therefore, you'll need to modify some settings upon each restart of the api server

marrouchi:
@yassine-sallemi Then I suggest we add onModuleInit for each module (HelperService, ChannelService), so each one cleans up his own settings.

api/src/plugins/plugins.service.ts
```diff
@@ -35,7 +44,37 @@ export class PluginService<T extends BasePlugin = BasePlugin> {
...
+
+    const orphanSettings = pluginSettings.filter(
+      (group) => !activePlugins.includes(group),
+    );
+
+    for (const group of orphanSettings) {
+      this.logger.log(`Deleting orphaned settings for ${group}...`);
+      await this.settingService.deleteGroup(group);
+    }
+  }
```
marrouchi, line 77:
```suggestion
  async cleanup(): Promise<void> {
  const activePlugins =this.getAll().map((handler) => handler.getNamespace());

  const pluginSettings = await this.settingService.getExtensionSettings('plugin');
  const orphanSettings = pluginSettings.filter(
    (namespace) => !activePlugins.includes(namespace)
  );

  await Promise.all(
    orphanSettings.map(async (group) => {
      this.logger.log(`Deleting orphaned settings for ${group}...`);
      return this.settingService.deleteGroup(group);
    })
  );
}
```

getExtensionSettings('plugin') would retrieve using a mongo query that searches for groups that end with "_plugin".

marrouchi (changes requested):

marrouchi:
Closed in favor of https://github.com/Hexastack/Hexabot/pull/883 

Thank you @yassine-sallemi for the effort you put in this PR.

PR #653, opened 2025-01-30
fix: resolve broadcastChannel duplication tab bug

### LINKED ISSUE #652: 🐛 [BUG] - broadcastChannel duplicate tab
### Description
We can notice that when we duplicate a tab the broadcastChannel is not working like expected

### Bug Demo
<video src="https://github.com/user-attachments/assets/b04c5a57-197d-4e2a-b016-bb73363ee751"></video>

### Proposed Solution
Removing unnecessary uuid logic is because broadcastChannel API is sending data only to other contexts. 

### Reproduction steps

```bash
1. Go to any page
2. Click on tab with right mouse button to click duplicate feature
3. Trigger a logout
4. See that the other tabs are not reloading
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation is to resolve duplication tab broadcastChannel bug.
The broadcastChannel API is sending postMessage message only to other contexts.

Fixes #652

image

Reference: 
https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API

# Type of change:

# Checklist:


marrouchi (changes requested):
Let's refactor the message / payload typing then.

yassinedorbozgithub:
> Let's refactor the message / payload typing then.

100% happy to do that

marrouchi (approved):

PR #655, opened 2025-01-30
Fix: update accepted attachments mime types

### LINKED ISSUE #650: Block/Content Attachment Uploader doesn't accept all files.

abdou6666:
# Motivation

This PR:
-  Adds new mimes types
 - Fix issue in the widget where we are not providing which mime types user can send
 - Fix issue in the user avatar where the input doesn't have accept values mime types thus able to pick a PDF as avatar
 
we should check if some of mime types i' have added aren't necessary

Fixes #650

# Type of change:

# Checklist:

### DIFF
--- a/frontend/src/app-components/inputs/FileInput.tsx
+++ b/frontend/src/app-components/inputs/FileInput.tsx
@@ -1,10 +1,11 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
  *
  * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
  * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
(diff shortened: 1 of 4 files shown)


yassinedorbozgithub (changes requested):

frontend/src/app-components/inputs/FileInput.tsx
```diff
@@ -71,6 +72,7 @@ const FileUploadButton = forwardRef<HTMLLabelElement, FileUploadButtonProps>(
           value="" // to trigger an automatic reset to allow the same file to be selected multiple times
           sx={{ display: "none" }}
           onChange={handleImportChange}
+          inputProps={{ accept }}
```
marrouchi, line 75: There's already a check up in this file, does it become unecessary?

marrouchi (changes requested):

frontend/src/app-components/inputs/FileInput.tsx
abdou6666: I think so with that attribute it won't display the options that aren't included in mime types

marrouchi (approved):

marrouchi (approved):

yassinedorbozgithub (approved):
LGTM :+1:

PR #657, opened 2025-01-30
feat: update docker-compose to use project name

abdou6666:
# Motivation

This PR will enable us to switch from different hexabot projects without the need to destroy volumes & lose db data.
image

Fixes # (issue)

# Type of change:

# Checklist:


docker/docker-compose.yml
```diff
@@ -23,7 +22,6 @@ services:
       start_period: 10s
 
   hexabot-frontend:
-    container_name: frontend
```
marrouchi, line 26: This will potential break the nginx config, we need to renae "hexabot-frontend" to "frontend"

docker/.env.example
```diff
@@ -66,3 +66,5 @@ APP_REDIS_PORT=9001
 REDIS_ENABLED=false
 REDIS_HOST=redis
 REDIS_PORT=6379
+
+COMPOSE_PROJECT_NAME=hexabot
```
marrouchi, line 70: Env var not in use.

marrouchi (changes requested):

docker/docker-compose.yml
abdou6666: should I keep the container names in that case?

PR #676, opened 2025-02-03
Feat: replace capture var to use zod

abdou6666:
# Motivation

This PR replaces joi with zod for `capture-var`

Fixes # (issue)

# Type of change:

# Checklist:


marrouchi (changes requested):
I think we will need to enforce the check on context_var : 
```
context_var: Joi.string()
        .regex(/^[a-z][a-z_0-9]*$/)
        .required(),
```

PR #677, opened 2025-02-03
Feat: add channel schema zod validation

abdou6666:
# Motivation

This PR adds schema validation for channel with `zod`
Fixes # (issue)

# Type of change:

# Checklist:

### DIFF
--- a/api/src/chat/schemas/types/channel.ts
+++ b/api/src/chat/schemas/types/channel.ts
@@ -1,10 +1,12 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
  *
  * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
  * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
  * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
  */
+
+import { z } from 'zod';
 
 import { ChannelName } from '@/channel/types';
 
@@ -17,3 +19,9 @@ export type SubscriberChannelData<C extends ChannelName = 'unknown-channel'> =
         // Channel's specific attributes
         [P in keyof SubscriberChannelDict[C]]: SubscriberChannelDict[C][P];
       };
+
+export const channelSchema = z.object({
+  name: z.string().regex(/-channel$/) as z.ZodType<ChannelName>,
+});
+
+export type Channel = z.infer<typeof channelSchema>;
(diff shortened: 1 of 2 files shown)


api/src/chat/validation-rules/is-position.ts
```diff
@@ -13,18 +13,10 @@ import {
...
 
-import { Position } from '../schemas/types/position';
+import { Position, positionSchema } from '../schemas/types/position';
 
 export function isPosition(position: Position) {
-  return (
-    typeof position === 'object' &&
-    !isNaN(position.x) &&
-    !isNaN(position.y) &&
-    position.x !== Infinity &&
```
abdou6666, line 23: @marrouchi I noticed in zod we have to specify `z.number().finite(),` if we don't want Infinity values should i make the corresponding changes for all numbers?

marrouchi: We can ignore the Infinity check part, z.number() suffice 🙏

api/src/chat/validation-rules/is-channel-data.ts
```diff
@@ -12,18 +12,21 @@ import {
...
+import { z } from 'zod';
 
-export function isChannelData(channel: any) {
-  return (
-    typeof channel === 'object' &&
-    channel.name &&
-    typeof channel.name === 'string'
-  );
+export const channelSchema = z.object({
+  name: z.string(),
```
marrouchi, line 18: I think we should do a regex check to ensure that the name ends with "-channel". WDYT?

marrouchi: Can we move the schema in the type file, do we have a "types/channel.ts" ?

marrouchi (changes requested):

api/src/chat/schemas/types/channel.ts
```diff
@@ -17,3 +19,9 @@ export type SubscriberChannelData<C extends ChannelName = 'unknown-channel'> =
         // Channel's specific attributes
         [P in keyof SubscriberChannelDict[C]]: SubscriberChannelDict[C][P];
       };
+
+export const channelSchema = z.object({
+  name: z.string().regex(/-channel$/) as z.ZodType<ChannelName>,
+});
+
+export type Channel = z.infer<typeof channelSchema>;
```
marrouchi, line 27:
```suggestion
export type ChannelData = z.infer<typeof channelSchema>;
```

api/src/chat/schemas/types/channel.ts
```diff
@@ -17,3 +19,9 @@ export type SubscriberChannelData<C extends ChannelName = 'unknown-channel'> =
         // Channel's specific attributes
         [P in keyof SubscriberChannelDict[C]]: SubscriberChannelDict[C][P];
       };
+
+export const channelSchema = z.object({
```
marrouchi, line 23:
```suggestion
export const channelDataSchema = z.object({
```

marrouchi (changes requested):

marrouchi (approved):

PR #690, opened 2025-02-05
refactor(frontend): update language dialogs

### LINKED ISSUE #687: Labels (add, edit, delete)

yassinedorbozgithub:
# Motivation
The main motivation is to refactor the Languages code base to use useDialogs hook instead of using useDialog hook.

Fixes #687

# Checklist:


frontend/src/components/languages/LanguageForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+    },
+    code: {
+      required: t("message.code_is_required"),
+    },
+  };
+  const onSubmitForm = (params: ILanguageAttributes) => {
+    if (data) {
+      updateLanguage({ id: data.id, params });
+    } else {
+      createLanguage(params);
```
IkbelTalebHssan, line 68:
```suggestion
data ? updateLanguage(payload) : createLanguage(payload);
```

frontend/src/components/languages/LanguageForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+    if (data) {
+      reset({
+        title: data.title,
+        code: data.code,
+        isRTL: data.isRTL,
+      });
+    } else {
+      reset();
+    }
+  }, [data, reset]);
```
abdou6666, line 82: is it a good idea to refactor this code into useResetForm hook ?

`
const useResetForm((resetFn,data) => {
  useEffect(() => {
    if (data) {
      resetFn(data);
    } else {
      resetFn();
    }
  }, [data, reset]);
})
`

IkbelTalebHssan (approved):
LGTM :rocket:

frontend/src/components/languages/LanguageForm.tsx
yassinedorbozgithub: Thank you for the suggestion, i think that's a good idea, but we need to make it after the refactoring otherwise the merged PR's will miss this proposed hook

frontend/src/components/languages/LanguageForm.tsx
yassinedorbozgithub: Thank you for the review, for a better readability we need to avoid ternary expressions except when the instructions has a return value.

frontend/src/components/languages/LanguageForm.tsx
marrouchi: @yassinedorbozgithub Let's open a issue about this.

marrouchi (approved):

frontend/src/components/languages/LanguageForm.tsx
yassinedorbozgithub: Hi @marrouchi, done https://github.com/Hexastack/Hexabot/issues/707
Thank you @abdou6666 for the idea :rocket:

PR #718, opened 2025-02-07
fix: use pages router instead of app router

abdou6666:
# Motivation

This PR removes the import & utilization of app router
Fixes # (issue)

# Type of change:

# Checklist:

### DIFF
--- a/frontend/src/hooks/useAuth.ts
+++ b/frontend/src/hooks/useAuth.ts
@@ -6,7 +6,6 @@
  * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
  */
 
-
 import { useRouter } from "next/router";
 import { useContext } from "react";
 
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useAuth.ts
```diff
@@ -33,12 +33,11 @@ export const useAuth = () => {
 
 export const useLogoutRedirection = () => {
   const router = useRouter();
-  const pathname = usePathname();
   const hasPublicPath = PUBLIC_PATHS.includes(router.pathname);
   const logoutRedirection = async (fullReload: boolean = false) => {
     if (!hasPublicPath) {
       const redirectUrl = `/${RouterType.LOGIN}?redirect=${encodeURIComponent(
-        pathname,
+        router.pathname,
```
yassinedorbozgithub, line 40: Hi @abdou6666, i think that the use of next/router will cause an issue especially for **Dynamic Routes**.
For example to to reproduce the bug
- Open Visual editor page 
- Create a new Flow and navigate to this flow
- Click logout button (notice that the ID is missed in the URL)
- Connect again 
- Notice that the redirection is not to the previous visual editor because the dynamic ID is missed

yassinedorbozgithub (changes requested):

yassinedorbozgithub (approved):

PR #753, opened 2025-02-14
fix: make SettingType.select inferring options as type

### LINKED ISSUE #752: 🤔 [ISSUE] - make SettingType.select return options type instead returning a string type
### Description

# Description
We can notice that currently the SettingType.select setting is having a string. The idea is to make it returning an inferring type from the options value 

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation is to make setting with SettingType.select type inferring the options types.

Fixes #752

# Checklist:

### DIFF
--- a/api/src/setting/index.d.ts
+++ b/api/src/setting/index.d.ts
@@ -1,10 +1,12 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
  *
  * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
  * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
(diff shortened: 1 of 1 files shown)


IkbelTalebHssan (approved):

abdou6666:
LGTM :rocket: 

image

abdou6666 (approved):

api/src/setting/index.d.ts
```diff
@@ -22,10 +24,14 @@ declare global {
             ? { [K in keyof T]: TNativeType<T[K]> }
             : T;
 
+  type SettingValue<K> = K['type'] extends SettingType.select
+    ? K['options'][number]
+    : TNativeType<K['value']>;
+
   type SettingObject<
```
abdou6666, line 31: can we export this type, few days ago while writing custom plugins. I wanted to create a private fields containing the args of the plugin and I couldn't type it

marrouchi: @abdou6666 The type is under "global", so it's accessible without the need to export nor import

Issue #707, opened 2025-02-06
🤔 [ISSUE] - Implement a useResetForm hook


yassinedorbozgithub:
### Description

# Description
The idea is to refactor all files to inlcude a new useResetForm hook 

Idea proposed by @abdou6666 

Exemple :
`const useResetForm((resetFn,data) => { useEffect(() => { if (data) { resetFn(data); } else { resetFn(); } }, [data, reset]); })`

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

PR #769, opened 2025-02-17
Refactor/custom hook use reset form

yassine-sallemi:
# Motivation

Code refactoring, custom hook useResetForm.
[Issue #707](https://github.com/Hexastack/Hexabot/issues/707)


IkbelTalebHssan (approved):
LGTM :rocket:

yassinedorbozgithub (changes requested):
Hi @yassine-sallemi, after the implementation of useDialogs hook i believe that we don't need to use reset function for forms of dialogs defaultValues can default the form with the correct values (creation/update)

yassinedorbozgithub (approved):

PR #848, opened 2025-03-23
feat: add unit tests global provider config

### LINKED ISSUE #847: 🤔 [ISSUE] - make LoggerService and EventEmitter2 global unit test providers
### Description

# Description
We can notice that most of the unit test needs the LoggerService as provider.
We can make easier the implementation by doing a global provider for all the unit tests  

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation is to make loggerService and eventEmitter2 as global providers.

Fixes #847

# Checklist:

### DIFF
--- a/api/src/utils/test/utils.ts
+++ b/api/src/utils/test/utils.ts
@@ -1,0 +1,53 @@
+/*
+ * Copyright © 2025 Hexastack. All rights reserved.
+ *
+ * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
+ * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
+ * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
+ */
+
+import { ModuleMetadata } from '@nestjs/common';
+import { EventEmitter2 } from '@nestjs/event-emitter';
+import { Test, TestingModule } from '@nestjs/testing';
+
+import { LoggerService } from '@/logger/logger.service';
+
+type TTypeOrToken = [
+  new (...args: any[]) => any,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  ...(new (...args: any[]) => any[]),
+];
+const findInstances = async <T extends TTypeOrToken>(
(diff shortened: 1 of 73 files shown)


api/src/utils/test/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  ...(new (...args: any[]) => any[]),
+];
+const findInstances = async <T extends TTypeOrToken>(
+  type: keyof TestingModule,
+  module: TestingModule,
+  typesOrTokens: T,
+): Promise<{ [K in keyof T]: InstanceType<T[K]> }> =>
+  Promise.all(
```
MohamedAliBouhaouala, line 26: I think it's safer to use a for loop in a testing context since most dependencies would be retrieved synchronously and we wouldn't need to introduce unnecessary overhead

api/src/utils/test/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+export const buildTestingMocks = async ({
+  providers,
+  ...rest
+}: ModuleMetadata) => {
+  const module = await Test.createTestingModule({
+    ...rest,
+    ...(providers && {
+      providers: [LoggerService, EventEmitter2, ...providers],
+    }),
+  }).compile();
```
MohamedAliBouhaouala, line 46: Just a quick question : Don't we need to use to use afterAll(() => {
  module.close();
}); in each test suite or is it handled differently (to avoid potential memory leaks) ? Please excuse my ignorance

IkbelTalebHssan (approved):

IkbelTalebHssan (approved):

api/src/utils/test/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  return {
+    module,
+    getMocks: extractInstances('get', module),
```
abdou6666, line 50: Suggestion to change `getMocks` to return an `object containing instances instead of an `array`, ensuring that destructuring order doesn't matter

abdou6666 (changes requested):

abdou6666 (approved):

MohamedAliBouhaouala (approved):

PR #880, opened 2025-04-03
fix: end conversation when a block have no matching outcomes

### LINKED ISSUE #878: 🤔 [ISSUE] - Bot hangs up when the last block has outcomes with no matching next block
### Description

Currently, when a conversation block in Hexabot has no matching outcomes, the conversation gets stuck, leaving the bot unresponsive. This issue occurs because the system expects every block to have at least one valid transition, and when none are found, the flow is interrupted without an explicit fallback or termination.

### Reproduction steps

```bash
1. Create a conversation flow where a block has multiple outcomes, but none match the user's input.
2. Trigger the conversation and provide input that does not correspond to any expected outcome.
3. Observe that the bot does not proceed further and remains in the current state indefinitely.
```

### Example

Image

### Logs

Image

IkbelTalebHssan:
# Motivation

addresses an issue in the Hexabot project where the bot would hang if a conversation block had no matching outcomes. The proposed change ensures that the conversation ends gracefully in such scenarios, preventing the bot from becoming unresponsive.

Fixes #878

# Type of change:

# Checklist:

### DIFF
--- a/api/src/chat/services/bot.service.ts
+++ b/api/src/chat/services/bot.service.ts
@@ -200,6 +200,7 @@ export class BotService {
                 'Block outcome did not match any of the next blocks',
                 convo,
               );
+              this.eventEmitter.emit('hook:conversation:end', convo, false);
             }
           } else {
             // Conversation continues : Go forward to next blocks


abdou6666:
LGTM! 

image

image

api/src/chat/services/bot.service.ts
```diff
@@ -200,6 +200,7 @@ export class BotService {
                 'Block outcome did not match any of the next blocks',
                 convo,
               );
+              this.eventEmitter.emit('hook:conversation:end', convo, false);
```
marrouchi, line 203: @IkbelTalebHssan The second parameter is passed on but not used by the listener. I think the initial idea is that it would indicate if the conversation has ended because of an error or not. I think it's safe to remove it from every `hook:conversation:end` emit. This can be the subject another PR,  can you open an issue about this ?

marrouchi (approved):

api/src/chat/services/bot.service.ts
IkbelTalebHssan: @marrouchi You're right :heavy_check_mark: We may also need to check this in other related projects.
I will create an issue and a PR related to this.
Thanks :pray:

PR #905, opened 2025-04-09
fix: train edit nlu sample

abdou6666:
# Motivation

Fixes # (issue)

# Type of change:

# Checklist:

### DIFF
--- a/api/src/nlp/services/nlp-sample.service.ts
+++ b/api/src/nlp/services/nlp-sample.service.ts
@@ -15,8 +15,11 @@ import { OnEvent } from '@nestjs/event-emitter';
 import Papa from 'papaparse';
 
 import { Message } from '@/chat/schemas/message.schema';
+import { HelperService } from '@/helper/helper.service';
+import { HelperType } from '@/helper/types';
 import { Language } from '@/i18n/schemas/language.schema';
 import { LanguageService } from '@/i18n/services/language.service';
+import { SettingService } from '@/setting/services/setting.service';
 import { BaseService } from '@/utils/generics/base-service';
 import { THydratedDocument } from '@/utils/types/filter.types';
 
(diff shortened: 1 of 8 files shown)


api/src/nlp/services/nlp-sample.service.ts
```diff
@@ -304,4 +309,34 @@ export class NlpSampleService extends BaseService<
...
+        type: 'inbox',
+      });
+      if (!foundSample) {
+        return;
+      }
+      await this.nlpSampleEntityService.storeSampleEntities(
+        foundSample,
+        entities,
+      );
+      await this.repository.updateOne(foundSample.id, {
```
MohamedAliBouhaouala, line 337: Language should also be updated because it's actually taking the default value.

MohamedAliBouhaouala:
We need to handle the migration of newly added settings. Currently it's not working as far as I tested.

abdou6666:
@MohamedAliBouhaouala Thanks for the review, feed back applied !

api/src/nlp/services/nlp-sample.service.ts
```diff
@@ -330,12 +336,25 @@ export class NlpSampleService extends BaseService<
       if (!foundSample) {
         return;
       }
+
       await this.nlpSampleEntityService.storeSampleEntities(
         foundSample,
         entities,
       );
+
+      const language = await this.languageService.findOne(
```
MohamedAliBouhaouala, line 345: Just minor enhancement to avoid extra DB calls: 

 let languageId: Language | null = null;
      if (languageEntity?.value) {
        languageId = await this.languageService.findOne(
          { code: languageEntity.value },
          undefined,
          { _id: 1 }
        );
      }

abdou6666:
This PR has been put on hold as other development tasks have taken higher priority

marrouchi:
We should emit 'hook:chatbot:received' after the NLU prediction, that way we avoid doing 2 requests.
```

      // Trigger message received event
      this.eventEmitter.emit('hook:chatbot:received', event);

      if (subscriber?.assignedTo) {
        this.logger.debug('Conversation taken over', subscriber.assignedTo);
        return;
      }

      if (event.getText() && !event.getNLP()) {
        try {
          const helper = await this.helperService.getDefaultNluHelper();
          const nlp = await helper.predict(event.getText(), true);
          event.setNLP(nlp);
        } catch (err) {
          this.logger.error('Unable to perform NLP parse', err);
        }
      }
```

Issue #858, opened 2025-03-24
🤔 [ISSUE] -  Remove Obsolete Commented Code and Comments


MohamedAliBouhaouala:
### Description

There is commented-out code and outdated comments present in the codebase. These are no longer needed and may cause confusion for future development. [

![Image](https://github.com/user-attachments/assets/f5e6bdef-1905-4cae-a497-4f4f66f1694f)

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

yassinedorbozgithub:
The version 2 is no more supported

PR #921, opened 2025-04-13
feat: add contextVar regex validation

### LINKED ISSUE #917: 💡 [REQUEST] - Add validation rules to context variables
### Summary

This issue proposes adding functionality to validate context variables based on user-defined rules. Currently, there isn't a flexible way to ensure that context variables conform to specific formats or patterns. This feature would empower users to define custom validation logic, enhancing data integrity and preventing unexpected behavior in downstream processes.

**Motivation:**

In various scenarios, it's crucial to ensure that context variables adhere to specific formats. For example:

* Ensuring an order ID context variable only contains digits.
* Verifying that a tracking number follows a specific alphanumeric pattern.
* Validating that a user input matches a defined regular expression.

Without built-in support for such validation, users often have to implement these checks manually within their workflows, leading to code duplication and potential inconsistencies.

**Proposed Solution:**

We propose introducing a mechanism that allows users to define validation rules for context variables. This could be implemented in several ways, including:

* **Simple Rule Types:** Providing predefined rule types such as "digits only," "alphanumeric," "minimum length," "maximum length," etc.
* **Pattern Matching:** Allowing users to specify a digit pattern (e.g., `DDD-DDDD-DDD`).
* **Regular Expression Support:** Enabling users to define validation rules using regular expressions for more complex patterns.
* **Custom Validation Functions (Advanced):** For highly specific requirements, allowing users to provide a custom function that takes the context variable as input and returns a boolean indicating validity.

### Basic Example

**Example Usage:**

Imagine a workflow where a user provides a "zip\_code" as a context variable. The user could define a validation rule like:

* **Rule Type:** `digits_only` and `length_equals: 5`
* **Pattern:** `\d{5}` (using a digit pattern)
* **Regex:** `^[0-9]{5}$` (using a regular expression)

If the "zip\_code" variable does not meet this rule, the system could:

* Raise an error and halt the workflow.
* Provide a warning or log message.
* Potentially trigger a fallback mechanism.

yassinedorbozgithub:
# Motivation:
The main motivation is to add an extra regex validation before updating contextVars (permanent, and not permanent).
This added feature will give us the possibility to use multiple contextVars in the same block without value interference. 

(NB: this update suppose that the regex for all the contextVars used in the same block are unique) 

Fixes #917

# Screenshots:
image
image

# Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced regex pattern support for context variable configurations. Users can now specify a custom regex pattern when creating or updating context variables.
  - Expanded the context variable form with a dedicated input field that validates the regex format.
  - Enhanced the display in context variable listings with a new column showing the configured regex pattern.

- **Refactor**
  - Improved validation logic ensures that only context variable values matching the defined patterns are accepted.
  - Updated method signatures to accommodate potential undefined values in context variable retrieval.

### DIFF
--- a/api/src/chat/services/conversation.service.ts
+++ b/api/src/chat/services/conversation.service.ts
@@ -116,9 +116,17 @@ export class ConversationService extends BaseService<
 
         const context_var = contextVars[capture.context_var];
         const { permanent, pattern = '' } = context_var || {};
-        const isValidContextValue = new RegExp(pattern.slice(1, -1)).test(
-          `${contextValue}`,
-        );
+        const isValidContextValue = (() => {
+          try {
+            return new RegExp(pattern.slice(1, -1)).test(`${contextValue}`);
+          } catch (error) {
+            this.logger.error(
+              `Invalid regex pattern for ContextVar[${capture.context_var}]: ${pattern}`,
+            );
+            return false;
+          }
+        })();
+
         if (isValidContextValue) {
           if (profile.context?.vars && permanent) {
             this.logger.debug(
--- a/api/src/chat/services/conversation.service.ts
+++ b/api/src/chat/services/conversation.service.ts
@@ -6,7 +6,7 @@
  * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
  */
 
-import { Injectable, Logger } from '@nestjs/common';
+import { Injectable } from '@nestjs/common';
 
 import EventWrapper from '@/channel/lib/EventWrapper';
 import { BaseService } from '@/utils/generics/base-service';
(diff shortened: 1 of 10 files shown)


api/src/chat/services/conversation.service.ts
```diff
@@ -114,16 +114,36 @@ export class ConversationService extends BaseService<
         contextValue =
           typeof contextValue === 'string' ? contextValue.trim() : contextValue;
 
-        if (
-          profile.context?.vars &&
-          contextVars[capture.context_var]?.permanent
-        ) {
-          Logger.debug(
-            `Adding context var to subscriber: ${capture.context_var} = ${contextValue}`,
+        const context_var = contextVars[capture.context_var];
```
MohamedAliBouhaouala, line 117: Lines 98-100.
There's obsolete commented out code. Can we get rid of it ?

abdou6666:
LGTM!
image

frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
```diff
@@ -26,7 +26,7 @@ import {
 import { OutcomeInput } from "./OutcomeInput";
 import { PostbackInput } from "./PostbackInput";
 
-const isRegex = (str: Pattern) => {
+export const isRegex = (str: Pattern) => {
```
abdou6666, line 29: We can do this instead to check for regex validation:
```
  try {
    new RegExp(str.slice(1, -1));
    return true;
  } catch (e) {
    return false;
  }
```

abdou6666 (changes requested):

api/src/chat/services/conversation.service.ts
yassinedorbozgithub: Thank you for the feedback, your are right but. 
Let's keep the update as short as possible to facilitate the review process.
I propose to create a dedicated [issue](https://github.com/Hexastack/Hexabot/issues/924) for the recommended update.

frontend/src/components/visual-editor/form/inputs/triggers/PatternInput.tsx
yassinedorbozgithub: The isRegex function is used to detect from an input the corresponding form type to show.
If we use this function the text case will be impacted directly.

api/src/chat/services/conversation.service.ts
MohamedAliBouhaouala: That's a duplicate issue. I already created it previously. See the [original one](https://github.com/Hexastack/Hexabot/issues/858)

abdou6666 (approved):

marrouchi:
Closed in favor of the new NLU pattern lookup strategy.

PR #923, opened 2025-04-13
fix(api): resolve messages order response

### LINKED ISSUE #906: 🐛 [BUG] - Incorrect Chatbot Response Sequencing
### Description

Analysis of the code in “bot.service.ts” (in particular the recursive triggerBlock method) shows that each time a block is “triggered,” the code calls sendMessageToSubscriber, which fires the "hook:chatbot:sent" event. That event, in turn, causes the message to be inserted into the database (via messageService) asynchronously. Meanwhile, triggerBlock immediately proceeds (or recurses) to any attached/next block if present.

Because these message insertions happen in parallel (the NestJS event emitter does not guarantee linear ordering or blocking), the second block’s “create” call in the database can sometimes finish before the first block’s insertion—even though the first block was conceptually triggered first. As a result, the createdAt timestamps in Mongo wind up “out of order,” with later-triggered blocks being written (or finishing the insert) sooner than earlier-triggered blocks.

### Reproduction steps

```bash
1. Open the 'Visual Editor'.
2. Add three 'Simple Text' blocks.
3. Connect the blocks sequentially using the attached link feature.
4. Repeatedly trigger the chatbot's entry point and observe the responses. Note that the order of the chatbot response messages sometimes differs from the expected linked sequence.
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation is to fix a bug impacting response messages via RestApi / Web Socket / long polling.

Fixes #906

# Screen recording:
## Reproduced bug :
- **Via Web Socket**
<video src="https://github.com/user-attachments/assets/45cc7ca5-802b-4bc5-bffe-3fc103f9bd28"></video>
- **Via Long Polling**
<video src="https://github.com/user-attachments/assets/d01d142b-0c6c-4a07-b01c-e6434a819ec6"></video>

# Type of change:

# Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Enhanced asynchronous processing for chatbot events to ensure smooth conversation flow and improved reliability.

### DIFF
--- a/api/src/chat/services/bot.service.ts
+++ b/api/src/chat/services/bot.service.ts
@@ -88,7 +88,7 @@ export class BotService {
       read: false,
       delivery: false,
     };
-    this.eventEmitter.emit('hook:chatbot:sent', sentMessage, event);
+    await this.eventEmitter.emitAsync('hook:chatbot:sent', sentMessage, event);
 
     // analytics log block or local fallback
     if (fallback) {
@@ -425,7 +425,12 @@ export class BotService {
           subscriber.id,
           block.name,
         );
-        return this.triggerBlock(event, updatedConversation, block, false);
+        return await this.triggerBlock(
+          event,
+          updatedConversation,
+          block,
+          false,
+        );
       } catch (err) {
         this.logger.error('Unable to store context data!', err);
         this.eventEmitter.emit('hook:conversation:end', convo);
--- a/api/src/chat/services/chat.service.ts
+++ b/api/src/chat/services/chat.service.ts
@@ -96,7 +96,7 @@ export class ChatService {
    *
    * @param sentMessage - The message that has been sent
    */
-  @OnEvent('hook:chatbot:sent')
+  @OnEvent('hook:chatbot:sent', { promisify: true })
   async handleSentMessage(
     sentMessage: MessageCreateDto,
     _event: EventWrapper<any, any>,


MohamedAliBouhaouala (approved):

abdou6666 (approved):

abdou6666:
LGTM!
image

IkbelTalebHssan (commented):
Hi @yassinedorbozgithub 
I couldn't reproduce the issue unfortunately but I encountered this misbehavior: when there is a block attached to a quick-replies one the options are hidden in the widget (which is not the case in the inbox interface) like shown in the screenshot below 
image

yassinedorbozgithub:
> when there is a block attached to a quick-replies one the options are hidden in the w

Hi @IkbelTalebHssan, yes this is correct.
I think that the build of flow needs to take in consideration the right (logical) order of the blocks, 
for example and following the mentioned flow, we can make the quickReplies (interactive block) the latest block by removing the next attached text block and the issue will be resolved

IkbelTalebHssan:
> > when there is a block attached to a quick-replies one the options are hidden in the w
> 
> Hi @IkbelTalebHssan, yes this is correct. I think that the build of flow needs to take in consideration the right (logical) order of the blocks, for example and following the mentioned flow, we can make the quickReplies (interactive block) the latest block by removing the next attached text block and the issue will be resolved

Yes @yassinedorbozgithub you're right. No need for an attached block if we're waiting for the user response/choice.

IkbelTalebHssan (approved):

PR #965, opened 2025-05-01
fix: resolve show more button issue

### LINKED ISSUE #964: 🤔 [ISSUE] - Carousel show more issue
### Description
The Carousel show more button (skip) is not working like expected. (When using outcome links) 

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation is to fix the show more button issue.

Fixes #964

# Type of change:

# Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved conversation flow handling to ensure context data is properly stored when progressing through system message outcomes.

### DIFF
--- a/api/src/chat/services/bot.service.ts
+++ b/api/src/chat/services/bot.service.ts
@@ -205,6 +205,14 @@ export class BotService {
                     block.attachedBlock,
                 );
               }
+
+              await this.conversationService.storeContextData(
+                convo,
+                nextBlock,
+                event,
+                false,
+              );
+
               return await this.triggerBlock(event, convo, nextBlock, fallback);
             } else {
               this.logger.warn(


api/src/chat/services/conversation.service.ts
```diff
@@ -156,7 +156,11 @@ export class ConversationService extends BaseService<
         next.options.content.display === OutgoingMessageFormat.carousel)
     ) {
       if (event.getPayload() === VIEW_MORE_PAYLOAD) {
-        convo.context.skip[next.id] += next.options.content.limit;
+        convo.context.skip[next.id] =
+          (!isNaN(convo.context.skip[next.id]) &&
+          typeof convo.context.skip[next.id] === 'number'
+            ? convo.context.skip[next.id]
+            : 0) + next.options.content.limit;
```
abdou6666, line 163: we aren't checking if limit is number i think it can be undefined

abdou6666 (changes requested):

api/src/chat/services/conversation.service.ts
abdou6666, line 163: i think we should refactor this code into a method that handles pagination in case of list or carousel.

abdou6666 (changes requested):

api/src/chat/services/conversation.service.ts
yassinedorbozgithub: The code is updated by reverting the old code and adding the **storeContextData** method when using outcomes links.

yassinedorbozgithub: The code is updated by reverting the old code and adding the **storeContextData** method when using outcomes links.

marrouchi (approved):

PR #970, opened 2025-05-01
Handle context update conflicts

marrouchi:
Reverts Hexastack/Hexabot#969

Concurrent update conflicts occur when two operations modify the same document (or nested subdocument) at the same time, potentially overwriting each other’s changes. In MongoDB, individual write operations are atomic at the document level (no partial writes), but multiple separate updates on the same document can interleave if not coordinated. Here’s how to prevent unintended overwrites:

## Use Targeted $set for Nested Fields

By default, doing a direct update of a subdocument in MongoDB (e.g., updating an object field by providing a new object) replaces the entire subdocument rather than merging fields. This can cause earlier changes to that subdocument to be lost. Solution: use $set with dot notation to update only the specific field path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved conversation flow handling to ensure context data is properly stored before moving to the next step when processing system messages.
- **New Features**
  - Added support for conditional flattening of update data in single and bulk update operations, enhancing data update flexibility.
- **Tests**
  - Introduced comprehensive tests for the new object flattening utility to ensure correct behavior across various scenarios.
- **New Utilities**
  - Added a utility function to flatten nested objects into dot-separated key-value pairs for streamlined data processing.
- **Enhancements**
  - Extended schema support to include dynamic fields allowing flexible key-value storage.

### DIFF
--- a/api/src/utils/helpers/flatten.spec.ts
+++ b/api/src/utils/helpers/flatten.spec.ts
@@ -76,6 +76,25 @@ describe('flatten', () => {
     });
   });
 
+  it('should handle arrays as values without recursing into them', () => {
+    const object = {
+      items: [1, 2, 3],
+      nested: {
+        moreItems: [4, 5, 6],
+        deepNested: {
+          evenMoreItems: [7, 8, 9],
+        },
+      },
+    };
+    const result = flatten(object);
+
+    expect(result).toStrictEqual({
+      items: [1, 2, 3],
+      'nested.moreItems': [4, 5, 6],
+      'nested.deepNested.evenMoreItems': [7, 8, 9],
+    });
+  });
+
   it('should support an object without nested object', () => {
     const object = {
       name: 'name',
--- a/api/src/utils/helpers/flatten.spec.ts
+++ b/api/src/utils/helpers/flatten.spec.ts
@@ -1,0 +1,95 @@
+/*
+ * Copyright © 2025 Hexastack. All rights reserved.
+ *
+ * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
+ * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
+ * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
+ */
+
+import { flatten } from './flatten';
+
+describe('flatten', () => {
+  it('should support a nested object with one nested level', () => {
+    const object = {
+      name: 'name',
+      context: { nestedField: 'value' },
+    };
+    const result = flatten(object);
+
+    expect(result).toStrictEqual({
+      name: 'name',
+      'context.nestedField': 'value',
+    });
+  });
+
+  it('should support a nested object with multiple nested levels', () => {
+    const object = {
+      name: 'name',
+      context: {
+        nestedField: 'value',
+        country: {
+          isoCode: 'usa',
+          phonePrefix: '+1',
+          countryName: 'United States',
+        },
+      },
+    };
+    const result = flatten(object);
+
+    expect(result).toStrictEqual({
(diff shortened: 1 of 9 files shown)


api/src/utils/helpers/flatten.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @returns A new object with flattened keys
+ */
+export const flatten = (
+  data: object,
+  prefix: string | undefined = undefined,
+  result: object = {},
+): object => {
+  if (Array.isArray(data)) {
+    throw new Error('Data should be an object!');
+  }
```
IkbelTalebHssan, line 23:
```suggestion
  if (data === null || Array.isArray(data)) {
    throw new Error('Input must be a non null and a non array');
  }
```

api/src/utils/helpers/flatten.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
+ * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
+ * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
+ */
+
+/**
+ * Flattens a nested object into a single-level object with dot-separated keys.
+ * @param data - The data object to flatten
+ * @param prefix - The optional base key to prefix to the current object's keys
+ * @param result - The optional accumulator for the flattened object  data
```
IkbelTalebHssan, line 13:
```suggestion
 * @param result - The optional accumulator for the flattened object  data
 * @throws If input is not a plain object (array or null).
```

api/src/utils/helpers/flatten.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @returns A new object with flattened keys
+ */
+export const flatten = (
```
marrouchi, line 16: Two distinct paths can collapse into the same dotted key if the original keys themselves contain dots ({ 'a.b': { c: 1 } }). Can you check ?

yassinedorbozgithub: Yes, i will check this use case :technologist:

api/src/utils/helpers/flatten.ts
yassinedorbozgithub: The data prop is typed with object type
image
The typescript will protect against this case (having a null value)
image

IkbelTalebHssan (commented):
@yassinedorbozgithub 
LGTM left a minor suggestion. 
One question please:
What would be the behavior in case we had a circular reference?

api/src/utils/helpers/flatten.ts
yassinedorbozgithub: The logic will return always the latest occurrence as final value 
`flatten({ 'a.b.c': 1, 'a.b': { c: 2 }, a: { b: { c: 3 } } });`
will return `{a.b.c: 3}` in this case

api/src/utils/helpers/flatten.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  it('should support an empty object', () => {
+    const result = flatten({});
+
+    expect(result).toStrictEqual({});
+  });
+
+  it('should throw an error if data is an array', () => {
+    expect(() => flatten([])).toThrow('Data should be an object!');
+  });
+});
```
abdou6666, line 118: I suggest we add a test case for this scenario where input is object with keys like this :
```
{
    'user.name': {
      id: 'Alice'},
    nested: {
      'country.name': 'France',
    },
  }
```

yassinedorbozgithub: The mentioned case is integrated to the unit tests cases
image

abdou6666 (approved):

IkbelTalebHssan (approved):

PR #1080, opened 2025-06-02
feat(api): add strict Setting types

### LINKED ISSUE #1042: 🐛 [BUG] - SettingType.multiple_attachment setting fields type
### Description

We can notice that for **SettingType.multiple_attachment** setting type the value need to have an array value.

Image

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation is to make setting types dynamic per item based on the type fields.

Fixes #1042

# Type of change:

# Checklist:

## Summary by CodeRabbit

- **Refactor**
	- Streamlined and unified type definitions for channel, helper, and plugin settings, improving consistency across the application.
- **Chores**
	- Updated copyright year to 2025.

### DIFF
--- a/api/src/setting/schemas/types.ts
+++ b/api/src/setting/schemas/types.ts
@@ -131,9 +131,9 @@ export type AnySetting =
 
 export type SettingDict = { [group: string]: Setting[] };
 
-export type StrictSetting<
-  U,
+export type ExtensionSetting<
   E extends object = object,
+  U extends AnySetting = AnySetting,
   K extends keyof BaseSchema = keyof BaseSchema,
 > = U extends any
   ? {
--- a/api/src/setting/schemas/types.ts
+++ b/api/src/setting/schemas/types.ts
@@ -6,6 +6,8 @@
  * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
  */
 
+import { BaseSchema } from '@/utils/generics/base-schema';
+
 import { Setting } from './setting.schema';
 
 export enum SettingType {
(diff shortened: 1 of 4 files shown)


api/src/setting/schemas/types.ts
```diff
@@ -128,3 +130,17 @@ export type AnySetting =
   | MultipleAttachmentSetting;
 
 export type SettingDict = { [group: string]: Setting[] };
+
+export type StrictSetting<
```
marrouchi:
```suggestion
export type ExtensionSetting<
```

api/src/setting/schemas/types.ts
```diff
@@ -128,3 +130,17 @@ export type AnySetting =
...
+  U,
```
marrouchi: Since we always use "AnySetting", I wonder if we just use "AnySetting" instead of U. Or should we enforce "U extends AnySetting"

yassinedorbozgithub: Yes, sure :technologist:

marrouchi (approved):

api/src/setting/schemas/types.ts
```diff
@@ -128,3 +130,17 @@ export type AnySetting =
   | MultipleAttachmentSetting;
 
 export type SettingDict = { [group: string]: Setting[] };
+
+export type ExtensionSetting<
```
medchedli, line 134:
```suggestion
/**
 * A generic type for creating extension-specific settings that combines base schema properties
 * with extension-specific properties.
 *
 * The type works by:
 * 1. Taking a base setting type (U) and excluding specified keys (K) from BaseSchema
 * 2. Merging the remaining properties with extension-specific properties (E)
 * 3. Filtering out any properties that are typed as 'never'
 *
 * @template E - Extension-specific properties that will be merged with the base setting type
 * @template U - The base setting type to extend from (defaults to AnySetting)
 * @template K - Keys from BaseSchema to exclude from the resulting type (defaults to all BaseSchema keys)
 */
export type ExtensionSetting<
```

PR #1083, opened 2025-06-02
fix(frontend): support contentType default name value

### LINKED ISSUE #1018: 🐛 [BUG] - Content Type Fields Form : Fields name do not reflect DB names
### Description

When content type is created programmatically, the UI regenerate field names in snake case based on the label, this may lead to data inconsistency.

### Reproduction steps

```bash
1. Create a content type programatically (field label = First Name and name = "firstName")
2. Open the content type form and check the field
3. You will find the field name as "first_name"
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation of this PR are : 
- [x] Make the API contentType field accepting name camelCase by updating the update DTO
- [x] Make the UI supporting default camelCase values

Fixes #1018

# Type of change:

# Checklist:

### DIFF
--- a/api/src/cms/dto/contentType.dto.ts
+++ b/api/src/cms/dto/contentType.dto.ts
@@ -14,7 +14,6 @@ import {
   IsNotEmpty,
   IsOptional,
   IsString,
-  Matches,
   Validate,
   ValidateNested,
 } from 'class-validator';
@@ -27,7 +26,6 @@ import { ValidateRequiredFields } from '../validators/validate-required-fields.v
 export class ContentField {
   @IsString()
   @IsNotEmpty()
-  @Matches(/^[a-z][a-z_0-9]*$/)
   name: string;
 
   @IsString()
--- a/frontend/src/components/content-types/ContentTypeForm.tsx
+++ b/frontend/src/components/content-types/ContentTypeForm.tsx
@@ -130,6 +130,8 @@ export const ContentTypeForm: FC<ComponentFormProps<IContentType>> = ({
               gap={2}
             >
               <FieldInput
+                defaultLabel={contentType?.fields?.[index]?.label}
+                defaultName={contentType?.fields?.[index]?.name}
                 setValue={setValue}
                 control={control}
                 remove={remove}
--- a/frontend/src/components/content-types/components/FieldInput.tsx
+++ b/frontend/src/components/content-types/components/FieldInput.tsx
@@ -26,6 +26,8 @@ import { slugify } from "@/utils/string";
 export const FieldInput = ({
   setValue,
   index,
+  defaultLabel,
+  defaultName,
   ...props
 }: {
   index: number;
@@ -33,6 +35,8 @@ export const FieldInput = ({
   remove: UseFieldArrayRemove;
   control: Control<Partial<IContentType>>;
   setValue: UseFormSetValue<Partial<IContentType>>;
+  defaultLabel?: string;
+  defaultName?: string;
 }) => {
   const { t } = useTranslate();
   const label = useWatch({
@@ -41,7 +45,11 @@ export const FieldInput = ({
   });
 
   useEffect(() => {
-    setValue(`fields.${index}.name`, label ? slugify(label) : "");
+    if (defaultLabel && defaultName !== slugify(defaultLabel)) {
+      defaultName && setValue(`fields.${index}.name`, defaultName);
+    } else {
+      setValue(`fields.${index}.name`, label ? slugify(label) : "");
+    }
   }, [label, setValue, index]);
 
   return (


frontend/src/components/content-types/components/FieldInput.tsx
```diff
@@ -41,7 +45,11 @@ export const FieldInput = ({
   });
 
   useEffect(() => {
-    setValue(`fields.${index}.name`, label ? slugify(label) : "");
+    if (defaultLabel && defaultName !== slugify(defaultLabel)) {
```
marrouchi: Let's avoid using useEffect(), should rather update the name when the label is updated.

yassinedorbozgithub: Sure i will update the logic, apply the enhancements and push an update

frontend/src/components/content-types/components/FieldInput.tsx
```diff
@@ -67,6 +59,24 @@ export const FieldInput = ({
             label={t("label.label")}
             error={!!fieldState.error}
             helperText={fieldState.error?.message}
+            onChange={(e) => {
+              const currentValue = e.target.value;
+              const { label, name } =
+                props.control._defaultValues.fields?.find(
+                  (field) => field?.uuid === uuid,
+                ) || {};
```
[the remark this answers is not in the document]
yassinedorbozgithub: getValues don't exist under 'props.control'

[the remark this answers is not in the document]
yassinedorbozgithub: We need initial data not the current data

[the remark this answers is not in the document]
yassinedorbozgithub: fields will not reflect the API response but the current state of the component.
And in this situation we need the API response data.

frontend/src/components/content-types/ContentTypeForm.tsx
```diff
@@ -19,34 +19,71 @@ import { useToast } from "@/hooks/useToast";
...
+      name: "",
+      fields: FIELDS_FORM_DEFAULT_VALUES,
     },
   });
   const { append, fields, remove } = useFieldArray({
     name: "fields",
+    rules: {
+      validate: (value) => {
+        const labelCounts = value.reduce((acc, field) => {
+          if (!field.label.trim()) return acc;
```
medchedli: I think that using the name for validation instead of the label is actually a better approach in this case.

medchedli: The current validation only checks for duplicate labels, but it's case-sensitive. This means that for example "First Name" and "first name" are treated as different labels

frontend/src/components/content-types/ContentTypeForm.tsx
```diff
@@ -143,7 +158,11 @@ export const ContentTypeForm: FC<ComponentFormProps<IContentType>> = ({
               startIcon={<AddIcon />}
               variant="contained"
               onClick={() =>
-                append({ label: "", name: "", type: ContentFieldType.TEXT })
+                append({
+                  label: "",
+                  name: "",
+                  type: ContentFieldType.TEXT,
```
medchedli, line 138: I suggest we use `generateId()` to generate an uuid for newly appended fields.
```suggestion
                  type: ContentFieldType.TEXT,
                  uuid: generateId(),
```

api/src/cms/utilities/field-validation.utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  fieldName: keyof T,
+): boolean => {
+  if (!Array.isArray(fields)) return false;
+  const seen = new Set<string>();
+  return fields.every((f) => {
+    const fieldValue = f[fieldName] as string;
+    if (seen.has(fieldValue)) return false;
+    seen.add(fieldValue);
+    return true;
+  });
```
marrouchi:
```suggestion
  return new Set(fields.map(f => f[fieldName] as string)).size === fields.length;
```

marrouchi: We can apply the same in UI part ?

api/src/cms/dto/contentType.dto.ts
```diff
@@ -58,6 +57,7 @@ export class ContentTypeCreateDto {
   @ValidateNested({ each: true })
   @Validate(ValidateRequiredFields)
   @Type(() => ContentField)
+  @UniqueFieldNames()
```
MohamedAliBouhaouala, line 60: This was made in order to ensure that a content field has a unique combination, right?

api/src/cms/utilities/field-validation.utils.ts
yassinedorbozgithub: Sure

api/src/cms/dto/contentType.dto.ts
yassinedorbozgithub: Yes exactly :100:

frontend/src/components/content-types/ContentTypeForm.tsx
```diff
@@ -33,20 +38,34 @@ export const ContentTypeForm: FC<ComponentFormProps<IContentType>> = ({
...
     formState: { errors },
     handleSubmit,
-  } = useForm<Partial<IContentType>>({
-    defaultValues: {
-      name: contentType?.name || "",
-      fields: contentType?.fields || FIELDS_FORM_DEFAULT_VALUES,
+  } = useForm<IContentType>({
+    defaultValues: contentType || {
+      name: "",
+      fields: FIELDS_FORM_DEFAULT_VALUES,
```
marrouchi: I think contentType may contain additional attributes that are not relevant to the form (example : createdAt, updatedAt, ...)

frontend/src/components/content-types/components/FieldInput.tsx
```diff
@@ -8,82 +8,70 @@
...
         render={({ field, fieldState }) => (
           <Input
-            disabled={props.disabled}
+            disabled={isDisabled}
             {...field}
             label={t("label.label")}
             error={!!fieldState.error}
             helperText={fieldState.error?.message}
+            onChange={(e) => {
+              props?.onLabelChange?.(e.target.value);
```
marrouchi: I wonder if it's better to remove onLabelChange() and move the name update logic here ? WDYT ?

marrouchi (approved):
Looks neat ! 👏  Left some comments, I'm not sure if it's relevant though ...

frontend/src/components/content-types/components/FieldInput.tsx
yassinedorbozgithub: I will test the suggestions and keep you informed :technologist:

yassinedorbozgithub: Hi @marrouchi, updates are pushed :white_check_mark:

MohamedAliBouhaouala (approved):
LGTM 👍
Screenshot from 2025-06-11 11-24-46

PR #1089, opened 2025-06-03
feat(inbox): Only update subscriber cache when no filters are active

medchedli:
This PR modifies the subscriber cache update behavior in the inbox component to prevent unwanted updates when filters are applied. Previously, new subscribers were always added to the cache regardless of active filters, which could interfere with search results and filtered views.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved subscriber list updates to ensure new subscribers appear correctly in the unfiltered view.

### DIFF
--- a/frontend/src/components/inbox/hooks/useInfiniteLiveSubscribers.ts
+++ b/frontend/src/components/inbox/hooks/useInfiniteLiveSubscribers.ts
@@ -73,8 +73,9 @@ export const useInfiniteLiveSubscribers = (props: {
       if (event.op === "newSubscriber") {
         const { result } = normalizeAndCache(event.profile);
 
+        // Only update the unfiltered (all-subscribers) cache
         queryClient.setQueryData(
-          [QueryType.infinite, EntityType.SUBSCRIBER, params],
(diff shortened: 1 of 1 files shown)


frontend/src/components/inbox/hooks/useInfiniteLiveSubscribers.ts
```diff
@@ -72,22 +72,27 @@ export const useInfiniteLiveSubscribers = (props: {
...
 
-        queryClient.setQueryData(
-          [QueryType.infinite, EntityType.SUBSCRIBER, params],
-          (oldData) => {
-            if (oldData) {
-              const data = oldData as InfiniteData<string[]>;
+        // Only update cache if no filters are applied
+        if (!hasFilters) {
+          queryClient.setQueryData(
+            [QueryType.infinite, EntityType.SUBSCRIBER, params],
```
marrouchi:
```suggestion
            [QueryType.infinite, EntityType.SUBSCRIBER, { where: {} }],
```
We can set the query data regardless (by removing the if condition)

marrouchi (approved):

PR #1093, opened 2025-06-03
Add type checks for string inputs in formatWithSlashes and extractRegexBody functions

medchedli:
Implement type checks to ensure that inputs for the `formatWithSlashes` and `extractRegexBody` functions are strings, enhancing input validation and preventing potential errors.

These changes add proper type guards to prevent the startsWith error by:
1. Adding a type check in formatWithSlashes to return a default value if the input is not a string
2. Adding a type check in extractRegexBody to return an empty string if the input is not a string

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of unexpected input types in string formatting utilities, resulting in more robust and predictable behavior for end-users.


frontend/src/utils/string.ts
```diff
@@ -31,6 +31,7 @@ export const isRegexString = (str: any) => {
  */
 export const formatWithSlashes = (value: string): string => {
   if (!value) return "/";
+  if (typeof value !== "string") return "/";
```
marrouchi:
```suggestion
  if (!value || typeof value !== "string") return "/";
```
Let's check why is it a single "/"

frontend/src/utils/string.ts
```diff
@@ -40,12 +41,13 @@ export const formatWithSlashes = (value: string): string => {
 /**
  * Extracts the inner regex from /.../
  */
-export const extractRegexBody = (value: string | undefined): string => {
+export const extractRegexBody = (value: string | undefined): string => {
+  if (typeof value !== "string") return "";
   if (value && value.startsWith("/") && value.endsWith("/")) {
```
marrouchi:
```suggestion
  if (value && typeof value === 'string' && value.startsWith("/") && value.endsWith("/")) {
```

yassinedorbozgithub:
From my understanding the root cause of the issue is that the value is not reflecting the current value of the form.
I recommend to use watch function to get the value with the index (idx)

marrouchi (approved):

frontend/src/utils/string.ts
```diff
@@ -30,7 +30,7 @@ export const isRegexString = (str: any) => {
  * Ensures value is wrapped in slashes: /value/
  */
 export const formatWithSlashes = (value: string): string => {
-  if (!value) return "/";
+  if (!value || typeof value !== "string") return "//";
```
MohamedAliBouhaouala, line 33: Screenshot from 2025-06-09 11-32-12  
exception stems from the fact that we were performing regex match against an object 👍

frontend/src/utils/string.ts
```diff
@@ -30,7 +30,7 @@ export const isRegexString = (str: any) => {
  * Ensures value is wrapped in slashes: /value/
  */
 export const formatWithSlashes = (value: string): string => {
```
MohamedAliBouhaouala, line 32: value can be an object too

medchedli: Yes, this point was addressed in this PR: https://github.com/Hexastack/Hexabot/pull/1111. We refactored `TriggersForm` to use `useFieldArray` from react-hook-form, which ensures the correct values are passed for each pattern.

marrouchi (approved):

Issue #1099, opened 2025-06-04
Need to define indexes for schemas


marrouchi:
Docs: https://wanago.io/2021/09/20/api-nestjs-indexes-mongodb-mongoose/

To avoid performance degradation on large datasets, please add indexes in your Nlp schemas under api/src/nlp/schemas:

Example: In NlpSampleEntity schema (e.g. nlp-sample-entity.schema.ts):
- index entity and value fields for the initial $match
- index sample field to optimize the $lookup join

In NlpSample schema (e.g. nlp-sample.schema.ts):
- index _id is implicit, but ensure it’s not disabled
- index language (and any other filter fields) used in pipeline filters

We can use Mongoose decorators or schema-level indexes, for example:

```ts
@Schema()
@Index({ entity: 1, value: 1, sample: 1 })
export class NlpSampleEntity { … }

const NlpSampleSchema = SchemaFactory.createForClass(NlpSample);
NlpSampleSchema.index({ language: 1 });
```

marrouchi:
We need to define the following indices (individual, not combined) : 
- BotStats: type, day
- Attachment : resourceRef
- Block : name
- Category : label
- Label: title, label
- Subscriber : first_name, last_name, foreign_id
- ContentType : name
- Content : title
- Setting : group, label
- Permission : action, role, model

We need to also check in depth the configurations in Nestjs decorators and mongoose (autoIndex, autoSearchIndex, excludeIndexes, ..) in order to understand what to enable and when (on which environement : process.env.NODE_ENV).

PR #1138, opened 2025-06-16
feat: add schema indexes

MohamedAliBouhaouala:
# Motivation

The following PR introduces new indexes to the database schemas to optimize query performance and improve data retrieval efficiency. The added indexes target frequently queried fields identified through analysis of query patterns and slow queries, resulting in faster lookup times and reduced database load.
It is worth noting that no data migrations are necessary since indexes can be created on existing collections without modifying the data when applying indexes on an existing dump.

Fixes # (1099)[https://github.com/Hexastack/Hexabot/issues/1099]

# Type of change:

# Checklist:

### DIFF
--- a/api/src/chat/schemas/subscriber.schema.ts
+++ b/api/src/chat/schemas/subscriber.schema.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
  *
  * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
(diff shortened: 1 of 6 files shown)


api/src/chat/schemas/subscriber.schema.ts
```diff
@@ -154,7 +157,10 @@ export type SubscriberDocument = THydratedDocument<Subscriber>;
 
 export const SubscriberModel: ModelDefinition = LifecycleHookManager.attach({
   name: Subscriber.name,
-  schema: SchemaFactory.createForClass(SubscriberStub),
+  schema: SchemaFactory.createForClass(SubscriberStub).index({
+    first_name: 1,
+    last_name: 1,
+  }),
```
marrouchi, line 162: Can we use the decorator at the class level ?

PR #1155, opened 2025-06-22
fix(api): enhance projection parameters typing

### LINKED ISSUE #1154: 🤔 [ISSUE] - enhance projection parameters typing
### Description

We can notice that for projections the current type allow to put any value

Screenshot:
Image

Example
```
    const deletedLanguages = await this.languageService.find(
      criteria,
      undefined,
      {
        id: 'test',
      },
    );
```

Expected behavior:
Allows only 1 or 0 as values

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivations of this PR is to enhance projection parameters typing.

Fixes #1154

# Screenshoot with the fix
image

# Checklist:

### DIFF
--- a/api/src/utils/types/filter.types.ts
+++ b/api/src/utils/types/filter.types.ts
@@ -146,3 +146,9 @@ export type THydratedDocument<T> = TOmitId<HydratedDocument<T>>;
 export type TFlattenOption = { shouldFlatten?: boolean };
 
 export type TQueryOptions<D> = (QueryOptions<D> & TFlattenOption) | null;
+
+export type TProjectField = 0 | 1;
+
+export type TProjectionType<T> = {
+  [K in keyof T]?: TProjectField;
+};
--- a/api/src/nlp/repositories/nlp-sample.repository.spec.ts
+++ b/api/src/nlp/repositories/nlp-sample.repository.spec.ts
@@ -219,12 +219,10 @@ describe('NlpSampleRepository', () => {
         skip: 0,
         sort: ['text', 'asc'],
       } as PageQueryDto<NlpSample>;
-      const projection = { text: 1 };
-
       const result = await nlpSampleRepository.findByEntitiesAndPopulate(
         { filters, values },
         page,
-        projection,
+        { text: 1 },
       );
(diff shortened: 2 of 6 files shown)


api/src/utils/types/filter.types.ts
```diff
@@ -146,3 +146,9 @@ export type THydratedDocument<T> = TOmitId<HydratedDocument<T>>;
 export type TFlattenOption = { shouldFlatten?: boolean };
 
 export type TQueryOptions<D> = (QueryOptions<D> & TFlattenOption) | null;
+
+export type TProjectField = 0 | 1;
+
+export type TProjectionType<T> = {
+  [K in keyof T]?: TProjectField;
+};
```
medchedli, line 154: According to MongoDB [Docs](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/crud/query/project/#projection-types), you cannot mix inclusion and exclusion in the same projection, except for the _id field. By default, _id is included, but you can explicitly exclude it. The type allows for example { status: 1, name: 0 }, which is invalid. We can fix in a new PR.

medchedli (approved):

api/src/utils/types/filter.types.ts
```diff
@@ -146,3 +146,9 @@ export type THydratedDocument<T> = TOmitId<HydratedDocument<T>>;
 export type TFlattenOption = { shouldFlatten?: boolean };
 
 export type TQueryOptions<D> = (QueryOptions<D> & TFlattenOption) | null;
+
+export type TProjectField = 0 | 1;
+
+export type TProjectionType<T> = {
```
medchedli, line 152:
```suggestion
/**
 * MongoDB projection type for selecting fields.
 * Use 1 to include a field, 0 to exclude.
 * Note: Do not mix inclusion and exclusion (except for _id).
 */
export type TProjectionType<T> = {
```

PR #1168, opened 2025-06-24
fix(api): add a TTL cache env variable

### LINKED ISSUE #1167: 🤔 [ISSUE] - add a cache TTL env variable
### Description

We can notice that we don't have an cache TTL env variable 

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation of this PR is to add a TTL cache env variable.

Fixes #1167

# Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring cache expiration duration using a new environment variable (`CACHE_TTL`). If not set, cache will default to a longer expiration time.
- **Documentation**
  - Updated environment variable example file to include the new `CACHE_TTL` setting.


api/src/config/index.ts
```diff
@@ -137,7 +137,7 @@ export const config: Config = {
   },
   cache: {
     type: process.env.REDIS_ENABLED === 'true' ? 'redis' : 'memory',
-    ttl: 60 * 1000, // Milliseconds
+    ttl: parseInt(process.env.CACHE_TTL || '3600000'), // Defaulted to 1 hour
```
marrouchi: Please let's set the default to 5 minutes

marrouchi (changes requested):

api/src/config/index.ts
yassinedorbozgithub: Happy to apply the required change :technologist:

marrouchi (approved):

PR #1178, opened 2025-06-27
fix(frontend): add support to target_block_id

### LINKED ISSUE #1177: 🤔 [ISSUE] - add support to target_block_id
### Description

We need to add a support to target_block_id

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation of this PR is to add support to target_block_id label.

Fixes #1177

# Checklist:

### DIFF
--- a/frontend/src/components/settings/SettingInput.tsx
+++ b/frontend/src/components/settings/SettingInput.tsx
@@ -153,6 +153,21 @@ const SettingInput: React.FC<RenderSettingInputProps> = ({
             {...rest}
           />
         );
+      } else if (setting.label === "target_block_id") {
+        const { onChange, ...rest } = field;
+
+        return (
+          <AutoCompleteEntitySelect<IBlock, "name", false>
+            searchFields={["name"]}
+            entity={EntityType.BLOCK}
+            format={Format.BASIC}
+            labelKey="name"
+            label={label}
+            multiple={false}
+            onChange={(_e, selected, ..._) => onChange(selected?.id || "")}
+            {...rest}
+          />
+        );
       }
 
       return (


frontend/src/components/settings/SettingInput.tsx
```diff
@@ -153,6 +153,21 @@ const SettingInput: React.FC<RenderSettingInputProps> = ({
             {...rest}
           />
         );
+      } else if (setting.label === "target_block_id") {
+        const { onChange, ...rest } = field;
+
+        return (
+          <AutoCompleteEntitySelect<IBlock, "name", false>
```
marrouchi: Let's group the options by flow/category please 🙏

marrouchi (changes requested):

frontend/src/components/settings/SettingInput.tsx
yassinedorbozgithub: Sure :technologist:

frontend/src/components/settings/SettingInput.tsx
```diff
@@ -153,6 +177,42 @@ const SettingInput: React.FC<RenderSettingInputProps> = ({
...
+                  color="primary"
+                >
+                  {group}
+                </Typography>
+                <Box>{children}</Box>
+              </li>
+            )}
+            filterOptions={(blocks) =>
+              blocks.filter(
+                (block) => block.message?.["plugin"] !== "redirect-plugin",
```
marrouchi: Avoiding an infinite loop is a condition that needs to be within the plugin itself, it's not the frontend responsability I think.

frontend/src/components/settings/SettingInput.tsx
```diff
@@ -153,6 +177,42 @@ const SettingInput: React.FC<RenderSettingInputProps> = ({
             {...rest}
           />
         );
+      } else if (setting.label === "target_block_id") {
```
marrouchi: For both "fallback_block" and "target_block_id", the setting could have any name. So we need to check if the setting is select and that the entity is "Block".

marrouchi (changes requested):

frontend/src/components/settings/SettingInput.tsx
```diff
@@ -115,19 +123,33 @@ const SettingInput: React.FC<RenderSettingInputProps> = ({
...
+                  p={2}
+                  fontWeight={700}
+                  color="primary"
+                >
+                  {group}
+                </Typography>
+                <Box>{children}</Box>
+              </li>
+            )}
+            {...config}
```
marrouchi: @yassinedorbozgithub Are you sure the "config" attributes are valid props to the component ?

marrouchi (changes requested):

frontend/src/components/settings/SettingInput.tsx
yassinedorbozgithub: I will double check

frontend/src/components/settings/SettingInput.tsx
```diff
@@ -115,19 +123,36 @@ const SettingInput: React.FC<RenderSettingInputProps> = ({
...
+                <Typography
+                  component="h4"
+                  p={2}
+                  fontWeight={700}
+                  color="primary"
+                >
+                  {group}
+                </Typography>
+                <Box>{children}</Box>
+              </li>
```
medchedli: I'm seeing a rendering problem with the list items. My guess is that renderGroup is using block names as keys, which leads to duplicate keys when several items have the same name. We'll need a unique identifier for each child.

medchedli (changes requested):

frontend/src/components/settings/SettingInput.tsx
yassinedorbozgithub: Hi @medchedli,
image
image
(key + li) warnings (console errors) are resolved :magic_wand:

marrouchi (approved):

PR #1197, opened 2025-07-04
fix(api): add Date support to flatten function

### LINKED ISSUE #1196: 🐛 [BUG] - Add Date support to flatten function
### Description

We can notice that the flatten function ignore Date (type object).
This behavior can cause issues.

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation of this PR are : 
- Add Date support to flatten function
- Add Unit test to cover the added logic

Fixes #1196 (HOTFIX)

# Type of change:

# Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Date objects when flattening nested objects, ensuring Date values are preserved correctly in the flattened output.

* **Tests**
  * Added a test to verify that objects containing Date instances are flattened without altering the Date values.

### DIFF
--- a/api/src/utils/helpers/flatten.ts
+++ b/api/src/utils/helpers/flatten.ts
@@ -10,7 +10,6 @@ const isPlainObject = (val: unknown): val is Record<string, unknown> => {
   return (
     typeof val === 'object' &&
     val !== null &&
-    !Array.isArray(val) && // not an Array
     Object.getPrototypeOf(val) === Object.prototype // direct Object.prototype
   );
 };
--- a/api/src/utils/helpers/flatten.ts
+++ b/api/src/utils/helpers/flatten.ts
@@ -6,6 +6,15 @@
  * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
  */
 
+const isPlainObject = (val: unknown): val is Record<string, unknown> => {
+  return (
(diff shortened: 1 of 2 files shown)


api/src/utils/helpers/flatten.ts
```diff
@@ -26,7 +26,12 @@ export const flatten = (
   for (const [key, value] of Object.entries(data)) {
     const path = prefix ? `${prefix}.${key}` : key;
 
-    if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
+    if (
+      typeof value === 'object' &&
+      !(value instanceof Date) &&
+      value !== null &&
+      !Array.isArray(value)
```
marrouchi: Let's declare the function that does test on plain objects:
```ts
const isPlainObject = (val: unknown): val is Record<string, unknown> =>
  typeof val === 'object' &&
  val !== null &&
  !Array.isArray(val) &&            // not an Array
  Object.getPrototypeOf(val) === Object.prototype;  // direct Object.prototype
```

then you do :
```ts
if (isPlainObject(value)) {
      flatten(value, path, result);          // keep drilling
    } else {
      result[path] = value;                  // treat everything else as leaf
    }
```

medchedli (approved):
:+1:

PR #1210, opened 2025-07-08
1209 issue   add authorization module

### LINKED ISSUE #1209: 🤔 [ISSUE] - Add authorization Module
### Description

We can centralize in one authorization Module all the REST/WebSocket access controls (based on user roles+permissions+models) 

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivationt
The main motivation of this PR is to add an Authorization Module that can manage access

Fixes #1209

# Checklist:

### DIFF
--- a/api/src/authorization/authorization.service.ts
+++ b/api/src/authorization/authorization.service.ts
@@ -1,0 +1,62 @@
+/*
+ * Copyright © 2025 Hexastack. All rights reserved.
+ *
+ * Licensed under the GNU Affero General Public License v3.0 (AGPLv3) with the following additional terms:
+ * 1. The name "Hexabot" is a trademark of Hexastack. You may not use this name in derivative works without express written permission.
+ * 2. All derivative works must include clear attribution to the original creator and software, Hexastack and Hexabot, in a prominent location (e.g., in the software's "About" section, documentation, and README file).
+ */
+
+import { Injectable } from '@nestjs/common';
+
+import { LoggerService } from '@/logger/logger.service';
+import { PermissionService } from '@/user/services/permission.service';
+import { UserService } from '@/user/services/user.service';
+import { MethodToAction } from '@/user/types/action.type';
+import { TModel } from '@/user/types/model.type';
+
+@Injectable()
+export class AuthorizationService {
+  constructor(
+    private readonly logger: LoggerService,
+    private readonly userService: UserService,
+    private readonly permissionService: PermissionService,
+  ) {}
+
+  /**
+   * Checks if a request (REST/WebSocket) is authorized to get access
+   *
+   * @param method - The Request Method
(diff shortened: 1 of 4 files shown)


api/src/user/guards/ability.guard.ts
```diff
@@ -69,26 +67,17 @@ export class Ability implements CanActivate {
...
-      } else {
+      try {
+        return await this.authorizationService.canAccess(
+          method,
+          user.id,
+          targetModel,
+        );
+      } catch (e) {
         throw new NotFoundException('Failed to load permissions');
       }
```
medchedli, line 82: I suggest we use throw `ForbiddenException` if canAccess returns false

```suggestion
      try {
        const accessGranted = await this.authorizationService.canAccess(
          method,
          user.id,
          targetModel,
        );
        if (!accessGranted) {
          throw new ForbiddenException(
            'You do not have permission to access this resource',
          );
        }
        return true;
      } catch (e) {
        throw new InternalServerErrorException(
          'An unexpected error occurred during authorization',
        );
      }
```

api/src/authorization/authorization.service.ts
```diff
@@ -0,0 +1,62 @@
...
+   * @param targetModel - The target model that we want access
+   * @returns
+   */
+  async canAccess(
+    method: string,
+    userId: string,
+    targetModel: TModel,
+  ): Promise<boolean> {
+    try {
+      const user = await this.userService.findOne(userId);
```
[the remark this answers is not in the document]
medchedli: We don't need to use findOneAndPopulate because we only use the role IDs for permission checks.

api/src/user/guards/ability.guard.ts
```diff
@@ -69,26 +67,17 @@ export class Ability implements CanActivate {
...
-        ?.split('/')[1]
-        .toLowerCase() as TModel | undefined;
-
-      const permissions = await this.permissionService.getPermissions();
-
-      if (permissions) {
-        const permissionsFromRoles = Object.entries(permissions)
-          .filter(([key, _]) => user.roles.includes(key))
-          .map(([_, value]) => value);
+      const targetModel = _parsedUrl?.pathname?.split('/')[1].toLowerCase() as
```
medchedli, line 70: I suggest we tackle this in a different PR, but it's worth mentioning that we can improve the curent implementation by using the built-in `@SetMetadata()` decorator from NestJS. This decorator will allow us to attach custom metadata to our controller routes to indicate the database model associated with the route and also the required user roles or permissions needed to access it. https://docs.nestjs.com/fundamentals/execution-context#reflection-and-metadata

yassinedorbozgithub:
No more relevant

PR #1211, opened 2025-07-08
feat: add schema indexes 

### LINKED ISSUE #1099: Need to define indexes for schemas

MohamedAliBouhaouala:
# Motivation
The following PR introduces new indexes to the database schemas to optimize query performance and improve data retrieval efficiency. The added indexes target frequently queried fields identified through analysis of query patterns and slow queries, resulting in faster lookup times and reduced database load.
It is worth noting that no data migrations are necessary since indexes can be created on existing collections without modifying the data when applying indexes on an existing dump. This is only valid `dev` environments. 
In production settings, `autoIndex` is disabled and an appropriate index migration script is needed

Fixes #1099

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved database performance by adding indexes to various fields across multiple schemas, including support for compound and text indexes.
  * Enhanced query efficiency for analytics, chat, attachment, NLP, and settings data.
  * Added sorting to analytics message queries for consistent result ordering.

* **Chores**
  * Updated copyright years to 2025.
  * Adjusted automatic index creation settings to enable it only outside of production environments.

### DIFF
--- a/api/src/nlp/schemas/nlp-sample.schema.ts
+++ b/api/src/nlp/schemas/nlp-sample.schema.ts
@@ -77,10 +77,7 @@ export class NlpSampleFull extends NlpSampleStub {
 export type NlpSampleDocument = THydratedDocument<NlpSample>;
 
 const NlpSampleSchema = SchemaFactory.createForClass(NlpSampleStub);
-NlpSampleSchema.index(
-  { text: 'text' },
-  { background: false, language_override: 'none' },
-);
+NlpSampleSchema.index({ text: 'text' }, { language_override: 'none' });
 
 export const NlpSampleModel: ModelDefinition = LifecycleHookManager.attach({
   name: NlpSample.name,
--- a/api/src/chat/schemas/subscriber.schema.ts
+++ b/api/src/chat/schemas/subscriber.schema.ts
@@ -94,6 +94,7 @@ export class SubscriberStub extends BaseSchema {
   @Prop({
     type: Date,
     default: () => Date.now(),
+    index: true,
   })
   lastvisit?: Date;
 
--- a/api/src/cms/schemas/content.schema.ts
+++ b/api/src/cms/schemas/content.schema.ts
@@ -101,7 +101,6 @@ ContentSchema.index(
       title: 2,
       rag: 1,
     },
-    background: false,
   },
 );
 
--- a/api/src/setting/schemas/setting.schema.ts
+++ b/api/src/setting/schemas/setting.schema.ts
@@ -58,6 +58,7 @@ export class Setting extends BaseSchema {
   @Prop({
     type: Number,
     default: 0,
+    index: true,
   })
   weight?: number;
 


api/src/utils/pipes/search-filter.pipe.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
```
yassinedorbozgithub:
```suggestion
 * Copyright © 2024 Hexastack. All rights reserved.
```

frontend/src/types/search.types.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Hexastack. All rights reserved.
+ * Copyright © 2025 Hexastack. All rights reserved.
```
yassinedorbozgithub:
```suggestion
 * Copyright © 2024 Hexastack. All rights reserved.
```

api/src/app.module.ts
```diff
@@ -63,6 +63,7 @@ const i18nOptions: I18nOptions = {
     MailerModule,
     MongooseModule.forRoot(config.mongo.uri, {
       dbName: config.mongo.dbName,
+      autoIndex: process.env.NODE_ENV !== 'production', // Disable in production
```
yassinedorbozgithub:
```suggestion
      autoIndex: config.env !== 'production', // Disable in production
```

yassinedorbozgithub (changes requested):
BotStats
image

frontend/src/types/search.types.ts
MohamedAliBouhaouala: Won't pass through linting @yassinedorbozgithub

api/src/app.module.ts
MohamedAliBouhaouala: Yup, nice catch

yassinedorbozgithub (changes requested):
Those two fields are candidate to be indexed (frequently used, configured as default sorting criteria)

- [Setting](http://localhost:8080/settings) page is using **weight** field (Number) as default sorting criteria
image


- [Inbox](http://localhost:8080/inbox) page is using **lastvisit** field (Date) as default sorting criteria
image

api/src/nlp/schemas/nlp-sample.schema.ts
```diff
@@ -76,12 +76,18 @@ export class NlpSampleFull extends NlpSampleStub {
 
 export type NlpSampleDocument = THydratedDocument<NlpSample>;
 
+const NlpSampleSchema = SchemaFactory.createForClass(NlpSampleStub);
+NlpSampleSchema.index(
+  { text: 'text' },
+  { background: false, language_override: 'none' },
```
marrouchi:
```suggestion
  { background: true, language_override: 'none' },
```

marrouchi (changes requested):

api/src/nlp/schemas/nlp-sample.schema.ts
medchedli: For reference, in MongoDB 7 and higher, index creation is mostly non-blocking, with only brief periods at the beginning and at the end of the process where the collection is locked. The explicit background option is deprecated, as the process is now optimized by default in newer versions of MongoDB. 

https://www.mongodb.com/docs/manual/core/index-creation/#comparison-to-foreground-and-background-builds

api/src/nlp/schemas/nlp-sample.schema.ts
```diff
@@ -76,12 +76,15 @@ export class NlpSampleFull extends NlpSampleStub {
 
 export type NlpSampleDocument = THydratedDocument<NlpSample>;
 
+const NlpSampleSchema = SchemaFactory.createForClass(NlpSampleStub);
```
medchedli, line 79: I suggest we also index the `language` field in NLP Sample

medchedli: These queries are not performance bottlenecks since the data volume is small :+1:

medchedli (approved):
LGTM

api/src/nlp/schemas/nlp-sample.schema.ts
MohamedAliBouhaouala: @medchedli  by default, it's true unless explicitly set otherwise.In older versions, it used to be false by default: 

https://docs.huihoo.com/mongodb/3.2/core/index-creation.html

PR #1233, opened 2025-07-14
fix(api): enhance uncaught exception handling

medchedli:
## Motivation
This change ensures that all critical errors, are always logged with a timestamp, even in the worst-case scenarios. Previously, the handler attempted to log errors asynchronously or in a way that could fail if the logger was unavailable (for example, during app startup or shutdown). This could result in **critical errors not being logged at all**, making post-mortem debugging very difficult.

To address this, the new implementation:
This change moves the global `uncaughtException` handler to use direct `ConsoleLogger` logging and removes all async operations (including dependency injection and LoggerService usage).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging for uncaught exceptions and unhandled promise rejections with detailed, timestamped information to enhance reliability during critical failures.

### DIFF
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -105,16 +105,16 @@ async function bootstrap() {
   }
 
   /**
-   * Global handler for uncaught exceptions.
+   * Global handler for uncaught exceptions in the Node.js process.
    *
-   * This handler logs all uncaught exceptions directly to stderr (console.error)
-   * with a timestamp and the error stack (or error object if no stack is present).
+   * Logs all uncaught exceptions directly to stderr using ConsoleLogger,
    *
-   * The log format is:
-   *   [timestamp] Uncaught Exception <error stack or error>
+   * Format:
+   *   [timestamp] Uncaught Exception <error stack | error message | error object>
    *
-   * NOTE: Do NOT include async calls inside this handler. The Node.js process may be in an unstable state,
-   * and async operations are discouraged and may not complete as expected.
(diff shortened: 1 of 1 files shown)


yassinedorbozgithub (changes requested):
Faced this when running the project with **EMAIL_SMTP_ENABLED**=true
image

medchedli:
> Faced this when running the project with **EMAIL_SMTP_ENABLED**=true image

I updated the handler to log directly to stderr. This approach guarantees that all fatal errors are reliably captured and that the process exits immediately and cleanly. To maintain service availability, this should be paired with a process manager like Docker or PM2 configured to automatically restart the service.

api/src/main.ts
```diff
@@ -105,11 +104,28 @@ async function bootstrap() {
...
+   *
+   * NOTE: Do NOT include async calls inside this handler. The Node.js process may be in an unstable state,
+   * and async operations are discouraged and may not complete as expected.
+   *
+   * @see https://nodejs.org/api/process.html#event-uncaughtexception
+   */
+  process.on('uncaughtException', (error) => {
+    const timestamp = new Date().toISOString();
+    // eslint-disable-next-line no-console
+    console.error(`[${timestamp}] Uncaught Exception`, error.stack || error);
```
marrouchi: LoggerService extends from ConsoleLogger, which is a class that you can instanciate directly. Can you try that ?

marrouchi (changes requested):

PR #1236, opened 2025-07-14
Flatten arrays in TranslationController for block and setting strings retrieval

medchedli:
Ensure that arrays are flattened when retrieving block strings and setting strings in the TranslationController. This improves the handling of string data.


yassinedorbozgithub (changes requested):
I suggest to to put the flat inside the two methods getAllBlockStrings, getSettingStrings

PR #1238, opened 2025-07-15
fix: Enhance DTO/Schema validations / UI Toast error presentation

### LINKED ISSUE #961: 🤔 [ISSUE] - Block update zod validation
## Description:
Currently, the updateBlock logic does not enforce strict validation on the input data. To ensure data integrity and catch issues early, we should introduce a zod schema for block updates.

Example : block options

## Scope
We are aligned to expand the scope of this issue after a Team discussion.

The points will cover : 

- Add a @Validate zod validation decorator in block.dto.ts, message.dto.ts, subscriber.dto.ts
- Refactor buildZodSchemaValidator function in goal to be used from the added decorator
- Remove validation-rules files
- Update useToast hook layout (to support grid layout)
- Remove t("message.internal_server_error") (error || t("message.internal_server_error") => error)
- Update useToast hook to support zod errors (array error?.message) 

## Proposed solution:

- Define a Zod schema that reflects the expected structure of a block during an update (e.g., name, type, transitions, settings, etc.).

- Use the schema inside the block service or controller to validate the input before processing the update.

- Return appropriate validation errors when the input does not conform to the schema.

## Benefits:

- Prevents malformed or incomplete block updates

- Makes the API behavior more predictable

- Improves type safety and maintainability

yassinedorbozgithub:
# Motivation
The main motivations of this PR are :
- [x] To add a @Validate zod validation decorator in block.dto.ts, message.dto.ts, subscriber.dto.ts
- [x] To refactor buildZodSchemaValidator function in goal to be used from the added decorator
- [x] To remove validation-rules files
- [x] To update useToast hook layout (to support grid layout)
- [x] To remove t("message.internal_server_error") (error || t("message.internal_server_error") => error)
- [x] To update useToast hook to support zod errors (array error?.message) 
- [x] To fix typing_indicator field sending type string -> number bug
- [x] To fix creating list bug
- [x] To fix plugin validation bug
- [x] To fix max fallback attempts bug
- [x] To remove number type from entity field (schema, unit tests, fixtures)
- [x] To enhance zod stringRegexPatternSchema to cover Empty regex, Invalid regex, Empty string cases.

Fixes #961

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved validation for form fields using schema-based decorators, resulting in more accurate and descriptive error messages.

* **Bug Fixes**
  * Error messages in toast notifications now display the full error object, improving error visibility in the frontend.
  * Validation for the "entity" field in content options now only accepts string or null values, ensuring consistency across the app.

* **Refactor**
  * Unified and simplified validation logic by replacing multiple custom decorators with a single schema-based approach.
  * Updated test and mock data to align with new validation requirements.


medchedli (approved):
🥇

yassinedorbozgithub:
- Creating list block issue
<video src="https://github.com/user-attachments/assets/6bc68c58-beda-4bbf-a5d7-b42df2831052"></video>

yassinedorbozgithub:
- Updating list block issue
<video src="https://github.com/user-attachments/assets/f91c4ae1-a9a6-48ed-9c89-6bebd59951a7"></video>

medchedli (changes requested):
Thank you, @yassinedorbozgithub. After a second round of testing, I was able to notice few the issue you highlighted in your videos. We'll need to address the problems with both the blockMessageObjectSchema and contentOptionsSchema validation.

Additionally, I noticed that the validator decorator applied to the message creation and subscriber creation DTOs is not currently enforced within their respective services. This means it's possible to bypass validation for channel data when creating a subscriber, and message length checks when creating a new message. I'm not certain how best to handle these cases, perhaps we can address them in a separate PR.

yassinedorbozgithub:
Plugin validation bug
<video src="https://github.com/user-attachments/assets/ace7d717-3466-4fd1-b102-53139a8a8b91"></video>

yassinedorbozgithub:
Hi @medchedli, new updates are pushed :technologist:

frontend/src/types/message.types.ts
```diff
@@ -76,7 +76,7 @@ export interface ContentOptions {
   buttons: AnyButton[];
   limit: number;
   query?: any; // Waterline model criteria
-  entity?: string | number | null; // ContentTypeID
+  entity?: string | null; // ContentTypeID
```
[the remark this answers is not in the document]
yassinedorbozgithub: Are you sure that this file api/src/nlp/schemas/nlp-sample-entity.schema.ts includes an entity field value and not an index

frontend/src/hooks/useToast.ts
```diff
@@ -42,12 +44,12 @@ const TOAST_WARNING_STYLE = {
...
   const { t } = useTranslation();
   const extractErrorMessage = (error: any) => {
-    if(typeof error === 'string') {
+    if (typeof error === "string") {
       return error;
-    }
-
-    else if (error?.statusCode == 409) {
+    } else if (error?.statusCode == 409) {
       return t("message.duplicate_error");
```
MohamedAliBouhaouala, line 50: But it can be a different error rather than `duplicate key error`, right?

frontend/src/hooks/useToast.ts
```diff
@@ -42,12 +44,12 @@ const TOAST_WARNING_STYLE = {
...
+    } else if (Array.isArray(error?.message)) {
+      return error?.message.toString();
```
MohamedAliBouhaouala, line 52: Is this type conversion always feasible?

frontend/src/hooks/useToast.ts
yassinedorbozgithub: Yes it will be handled by the next instructions
image

frontend/src/hooks/useToast.ts
yassinedorbozgithub: The zod validation can return multiple errors, for this case yes the .toString() method will be available if error?.message exist and it's an array

MohamedAliBouhaouala: But if it returns multiple errors, which error will be selected?

yassinedorbozgithub: All the returned errors will be showed in the UI

Issue #1251, opened 2025-07-21
💡 [REQUEST] - Display Date/Time for Messages in Inbox (Support “timeago” Format)


marrouchi:
### Implementation PR

Currently, messages displayed in the Inbox UI do not show any timestamp information, making it difficult for users to understand when a message was received or sent.

## Feature Request:

Please add support for displaying the date/time of each message in the inbox interface. Ideally:
Recent messages (within the past few hours/days) should use a “timeago” format, such as:
	•	2 minutes ago
	•	Yesterday at 3:15 PM
Older messages should fall back to a full date/time format, e.g.:
	•	June 10, 2025 at 14:32

### Summary

-

### Basic Example

-

### Drawbacks

-

PR #1264, opened 2025-07-22
feat: display timestamps for messages

MohamedAliBouhaouala:
# Motivation

The following PR enhances the Inbox UI by introducing timestamp visibility for each message, helping users better understand when messages were received or sent.

# Features
- **Timestamp Display**: Each message now shows a timestamp.
- **Relative "Timeago" Format**: Recent messages display times like:
  - `5 minutes ago`
  - `Yesterday 4:20 PM`
- **Full Date/Time Forma**t: Older messages fall back to:
  - `July 10, 2025 at 14:32`
Fixes # ([1251](https://github.com/Hexastack/Hexabot/issues/1251))

# Type of change:

# Checklist:

### DIFF
--- a/frontend/src/components/inbox/components/Chat.tsx
+++ b/frontend/src/components/inbox/components/Chat.tsx
@@ -25,6 +25,7 @@ import { EntityType } from "@/services/types";
 import { normalizeDate } from "@/utils/date";
 
 import {
+  formatSmartDate,
   getAvatarSrc,
   getMessageContent,
   getMessagePosition,
@@ -131,7 +132,10 @@ export function Chat() {
                         />,
                       ]
                     : []),
-                  ...getMessageContent(message),
+                  ...getMessageContent(
+                    message,
+                    formatSmartDate(message.createdAt),
+                  ),
                 ]}
               />
             );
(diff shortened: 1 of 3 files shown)


medchedli (changes requested):
It appears the `renderTimestamp()` is being called twice for messages that include quick replies or buttons. This is because these message types generate two separate message elements in the chat: one for the text content and another for the buttons. The same duplication occurs for attachments that have quick reply buttons. To avoid this redundancy, I suggest we only display the sent time for standard text messages.

medchedli (approved):

frontend/src/components/inbox/components/Chat.tsx
```diff
@@ -114,12 +114,6 @@ export function Chat() {
                     ? [
                         <Avatar
                           key={message.id}
-                          title={`${subscriber.first_name} ${
-                            subscriber.last_name
-                          } : ${normalizeDate(
-                            i18n.language,
-                            message.createdAt,
-                          )}`}
```
medchedli, line 122: I suggest keeping the name and removing only the date : 
```suggestion
                         title={`${subscriber.first_name} ${subscriber.last_name}`}
```

yassinedorbozgithub (approved):

PR #1270, opened 2025-07-29
fix(widget): resolve link hover colors

### LINKED ISSUE #1257: 🤔 [ISSUE] - Link Hover Color Matches Widget Theme, Causing Invisibility
### Description

When the widget theme is set to a color such as blue, links sent by users in the chat widget become invisible when hovered. This happens because the link hover color is set to match the widget's theme color, which is also used as the background for the message bubble. As a result, when a user hovers over a link, both the link text and the background are the same color, making the link unreadable.

### Reproduction steps

```bash
1. Set the widget theme color to blue.  
2. Open the chat widget and send a message containing a link (e.g., https://example.com).  
3. Hover over the link in the chat message.  
4. Observe that the link text color changes to blue, matching the background, and becomes invisible.
```

### Expected behavior:
The link should remain visible and accessible on hover, regardless of the widget theme color. The hover color for links should provide sufficient contrast with the background to ensure readability.

yassinedorbozgithub:
# What this PR does:
- Makes link colors change smoothly when hovering
- Keeps colors consistent across the site

# Why it matters:
🐛 Before: Some links has the same background color and the hover color
✅ Now: All links show the right hover color

# Changes:
- Fixed CSS for hover states
- Made colors match our design

Fixes #1257

# Type of change:

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Enhanced color themes with new hover color effects for sent and received messages, providing a more dynamic and interactive user experience.

* **Style**
  * Updated message background colors to use hover colors, improving visual feedback when interacting with messages.

### DIFF
--- a/widget/src/components/messages/TextMessage.tsx
+++ b/widget/src/components/messages/TextMessage.tsx
@@ -50,7 +50,7 @@ const TextMessage: React.FC<TextMessageProps> = ({ message }) => {
       style={{
         color: colors.text,
         backgroundColor: colors.bg,
-        ["--launcher-color" as string]: allColors.launcher.bg,
+        ["--launcher-color" as string]: colors.hover,
       }}
     >
       <p className="sc-message--text-content" ref={messageTextRef}>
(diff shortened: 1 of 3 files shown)


medchedli (approved):
👍

PR #1279, opened 2025-07-29
fix(api): resolve translation generation issue

### LINKED ISSUE #1242: 🤔 [ISSUE] - Translation generation bug when block setting changes from array of strings to single string
**Description:**

When modifying a custom block setting — for example, changing the `translatable` flag to `"true"` — this can implicitly change the data type of other settings (e.g., from `array of strings` to a `single string`). This change is not reflected or handled during translation key generation, leading to inconsistent or missing translation entries.

**Steps to Reproduce:**

1. Define a custom block with a setting that uses an array of strings (e.g., `["Option A", "Option B"]`).
2. Enable the `translatable` flag by setting it to `"true"`.
3. Update the setting value to a string instead of an array (e.g., `"Option A"`).
4. Trigger translation generation.

**Expected Behavior:**
Translation keys should be generated correctly regardless of whether the setting is a string or an array of strings.

**Actual Behavior:**
The translation generation logic fails or skips the value, assuming the original data structure (array), resulting in missing keys.

**Suggested Fix:**
Improve type-checking and fallback logic in the translation generation code to handle both `string` and `array` types gracefully.

**Impact:**
This may silently break multilingual support for blocks and result in incomplete translations in production.

yassinedorbozgithub:
# What this PR fixes:
- Makes translations work whether they expect strings or arrays
- Prevents errors when wrong type is provided

# Why it matters:
🐛 Before: Crashed if translation expected array but got string (or vice versa)
✅ Now: Handles both cases smoothly

# Changes made:
- Added type checks for translations
- Auto-convert between strings ↔ arrays when needed

Fixes #1242

# Type of change:

# Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of plugin settings to ensure correct processing of text values, with added safeguards and warnings for unexpected data types.
* **Chores**
  * Enhanced logging for better visibility into potential data type issues during translation processing.

### DIFF
--- a/api/src/i18n/services/translation.service.ts
+++ b/api/src/i18n/services/translation.service.ts
@@ -81,7 +81,7 @@ export class TranslationService extends BaseService<Translation> {
                 strings = strings.concat(value);
               } else if (typeof value === 'string') {
                 this.logger.warn(
-                  `The plugin ${plugin?.name} settings are incompatible with the settings.ts`,
+                  `The plugin ${plugin?.name} setting '${key}' is incompatible with the settings.ts`,
                 );
                 this.logger.warn(
                   `Expected type "array" received type "string"`,
@@ -95,14 +95,13 @@ export class TranslationService extends BaseService<Translation> {
                 strings.push(value);
               } else if (Array.isArray(value)) {
                 this.logger.warn(
-                  `The plugin ${plugin?.name} settings are incompatible with the settings.ts`,
+                  `The plugin ${plugin?.name} setting '${key}' is incompatible with the settings.ts`,
                 );
                 this.logger.warn(
                   'Expected type "string" received type "array"',
                 );
                 strings.push(...value.flat());
               }
-
               break;
             default:
               break;
--- a/api/src/i18n/services/translation.service.ts
+++ b/api/src/i18n/services/translation.service.ts
@@ -77,11 +77,32 @@ export class TranslationService extends BaseService<Translation> {
 
           switch (settingType) {
             case SettingType.multiple_text:
-              strings = strings.concat(value);
+              if (Array.isArray(value)) {
+                strings = strings.concat(value);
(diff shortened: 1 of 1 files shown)


api/src/i18n/services/translation.service.ts
```diff
@@ -77,11 +77,31 @@ export class TranslationService extends BaseService<Translation> {
...
+                strings = strings.concat(value);
+              } else if (typeof value === 'string') {
+                this.logger.warn(
+                  `The plugin ${plugin?.name} setting '${key}' is incompatible with the settings.ts`,
+                );
+                this.logger.warn(
+                  `Expected type "array" received type "string"`,
+                );
+                strings = strings.concat([value]);
+              }
```
marrouchi, line 94: Let's add an else and issue a warning if the type is different from the setting type.

api/src/i18n/services/translation.service.ts
```diff
@@ -77,11 +77,31 @@ export class TranslationService extends BaseService<Translation> {
...
+                strings.push(value);
+              } else if (Array.isArray(value)) {
+                this.logger.warn(
+                  `The plugin ${plugin?.name} setting '${key}' is incompatible with the settings.ts`,
+                );
+                this.logger.warn(
+                  'Expected type "string" received type "array"',
+                );
+                strings.push(...value.flat());
+              }
```
marrouchi, line 112: Same here

marrouchi (changes requested):

api/src/i18n/services/translation.service.ts
yassinedorbozgithub: Sure, happy to apply this :technologist:

api/src/i18n/services/translation.service.ts
yassinedorbozgithub: Sure

PR #1285, opened 2025-07-31
Feat/UI autocomplete entity select distinct component

### LINKED ISSUE #1278: 💡 [REQUEST] - Mutually Exclusive Labels for Subscribers
## Summary

Introduce a way to declare mutually exclusive labels so that a subscriber cannot hold two (or more) labels that conceptually conflict.
Example use case: subscription_tier labels — a subscriber may be one of free, basic, or premium, but never more than one at the same time.

This feature should be backward compatible (no change to the current Subscriber document shape) and easy to adopt incrementally.

## Problem Statement

Today, Subscriber.labels is a free-form array. Different flows may add labels such as free, basic, and premium without coordination, leading to inconsistent state and downstream logic branching on ambiguous inputs.

We need a declarative way to express “only one of these labels may exist at a time per subscriber”, enforced consistently and atomically across all write paths.

## Goals
- Keep Subscriber unchanged: labels: ObjectId[].
- Define and enforce mutual exclusivity between labels.
- Provide atomic updates to avoid race conditions (e.g., two processes assigning conflicting labels at the same time).
- Minimal migration and low risk.

## Non-Goals
- Changing how labels are stored on subscribers.
- Introducing a breaking API for existing label operations.
- Implementing full-blown rules engine beyond simple exclusivity.

## TODOs

- [x] Introduce API changes to add mutex group for labels
- [x] Update label form to be able to manage groups (created, assign, delete)
- [ ] Update the label dropdown input (subscriber > edit form, block form > options > assign labels) to be able to group labels and restrict end-user from selecting labels belonging to the same group.

yassinedorbozgithub:
**What this PR adds:**  
- A reusable component for grouped items  
- Lets users pick just one item per group  

**How it works:**  
☑️ Shows items in neat groups  
🔘 Only one selection allowed per group  

**Why it's useful:**  
✅ Solves our single-selection needs  
♻️ Works for any type of items  

# Screen recording 
<video src="https://github.com/user-attachments/assets/16f2dace-aedb-4bf2-ab1e-3dc7b6c2f024"></video>
<video src="https://github.com/user-attachments/assets/9106ea81-8da1-43a9-a8c1-da7066ec5ea4"></video>

Fixes #1278

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Introduced a new multi-select autocomplete component that allows users to select distinct entities grouped by a specified key, with improved grouping and sorting options.
  * Updated label selection in subscriber and visual editor forms to use the new grouped autocomplete, enhancing the organization and selection of labels.

* **Localization**
  * Added translations for "Default Group" in both English and French locales.

### DIFF
--- a/frontend/src/app-components/inputs/AutoCompleteEntityDistinctSelect.tsx
+++ b/frontend/src/app-components/inputs/AutoCompleteEntityDistinctSelect.tsx
@@ -8,15 +8,12 @@
 
 import { Box, ChipTypeMap, ListSubheader } from "@mui/material";
 import { AutocompleteProps } from "@mui/material/Autocomplete";
-import { forwardRef, useEffect, useRef } from "react";
+import { forwardRef } from "react";
 
+import { useFind } from "@/hooks/crud/useFind";
 import { useGetFromCache } from "@/hooks/crud/useGet";
-import { useInfiniteFind } from "@/hooks/crud/useInfiniteFind";
-import { useSearch } from "@/hooks/useSearch";
-import { Format, QueryType } from "@/services/types";
+import { Format } from "@/services/types";
 import { IEntityMapTypes, THook } from "@/types/base.types";
-import { TFilterStringFields } from "@/types/search.types";
-import { generateId } from "@/utils/generateId";
 
 import AutoCompleteSelect from "./AutoCompleteSelect";
 
@@ -37,7 +34,6 @@ type AutoCompleteEntityDistinctSelectProps<
   labelKey: Label;
   entity: keyof IEntityMapTypes;
   subEntity: keyof IEntityMapTypes;
-  searchFields: string[];
   disableSearch?: boolean;
   error?: boolean;
   helperText?: string | null | undefined;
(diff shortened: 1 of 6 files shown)


frontend/src/app-components/inputs/AutoCompleteEntityDistinctSelect.tsx
```diff
@@ -0,0 +1,186 @@
...
+      (a, idx, self) =>
+        self.findIndex((b) => a?.[idKey] === b?.[idKey]) === idx,
+    )
+    .sort((a, b) => (a[sortKey] ?? "").localeCompare(b[sortKey] ?? ""));
+  const options =
+    preprocess && flattenedData
+      ? preprocess((flattenedData || []) as unknown as Value[])
+      : ((flattenedData || []) as Value[]);
+
+  useEffect(() => {
```
medchedli: Isn't better to use memorized version of searchPayloadString ?
```suggestion
  const searchPayloadString = useMemo(
    () => JSON.stringify(searchPayload),
    [searchPayload],
  );

  useEffect(() => {
    fetchNextPage({ pageParam: params });
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [searchPayloadString]);
```

yassinedorbozgithub: Yes, but honestly i think that we don't need to search in this component

medchedli (approved):

PR #1303, opened 2025-08-05
Implement conversation search functionality with highlighting

medchedli:
Add conversation search capability, allowing users to search through chat history with highlighted results.


medchedli:
This proposal requires a significant change. We need to implement a backend search solution because the client-side approach has performance limitations.

PR #1313, opened 2025-08-11
feat: add profile and messages to settings WS response

### LINKED ISSUE #1312: 🤔 [ISSUE] - include profile and messages fields in the first WS response
### Description

To avoid extra subscribe requests on every refresh in the widget page we can includes the profile and the messages in the settings WS first response.

Old behavior :
- If we load the widget page 3 time we do three subscription WS requests 

New behavior :
- We do the subscription request only once (66% optimization of the request number)

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation of this PR is to avoid extra subscribe requests on every refresh in the widget page so we can include the profile and the messages in the settings WS first response.

# Old behavior :

If we load the widget page 3 times for example, three subscription WS `GET` requests are performed.

# New behavior :
 Subscription request is only performed once (66% optimization of the request number)

Fixes #1312

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Chat loads conversation history and subscriber profile on connect; launcher auto-opens when messages exist. New users with no history receive an automatic, localized "Get started" postback. Profile and initial messages are delivered to the widget.

* **Refactor**
  * Centralized message preprocessing to derive message direction and read/delivery semantics; chat context now exposes subscriber profile. Removed local profile persistence and automatic background subscribe-on-mount.

* **Chores**
  * Updated license year to 2025.

* **Quality**
  * Auto-scroll recalculates when message count changes; removed unused local-storage hook.


api/src/extensions/channels/web/base-web-channel.ts
```diff
@@ -131,9 +131,16 @@ export default abstract class BaseWebChannelHandler<
       try {
         const menu = await this.menuService.getTree();
         const hasSession = !!client.request.session.web?.profile;
+        const profile = client.request.session.web?.profile;
+        const messages = profile
+          ? await this.messageService.findHistoryUntilDate(profile)
+          : [];
+        const formattedMessages = await this.formatMessages(messages.reverse());
```
marrouchi: Let's add an "if" condition to check if the end-user has already submitted the form first. If not, we return undefined for profile and messages.

widget/src/providers/ChatProvider.tsx
```diff
@@ -399,9 +405,42 @@ const ChatProvider: React.FC<{
...
+          message.author === profile?.id
+            ? Direction.sent
+            : Direction.received;
+
+        message.direction = direction;
+        if (message.direction === Direction.sent) {
+          message.read = true;
+          message.delivery = false;
+        }
+      });
```
marrouchi: Let's refactor this into a helper function since this is performed twice. Right ?

yassinedorbozgithub: Yes, it's called multiple times

api/src/extensions/channels/web/base-web-channel.ts
```diff
@@ -130,10 +130,20 @@ export default abstract class BaseWebChannelHandler<
 
       try {
         const menu = await this.menuService.getTree();
-        const hasSession = !!client.request.session.web?.profile;
+        const profile = client.request.session.web?.profile;
+        let messages: Web.Message[] | undefined;
```
marrouchi:
```suggestion
        let messages: Web.Message[] | undefined = undefined;
```

widget/src/components/Launcher.tsx
```diff
@@ -43,6 +43,13 @@ const Launcher: React.FC<LauncherProps> = ({
 
   useSocketLifecycle();
 
+  useEffect(() => {
+    if (chat.messages.length > 0) {
+      widget.setScreen("chat");
+      chat.setConnectionState(3);
+    }
+  }, [chat, chat.messages.length, widget]);
```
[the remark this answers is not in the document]
marrouchi: I think this comment is very relevant.

widget/src/providers/ChatProvider.tsx
```diff
@@ -46,6 +48,40 @@ export const getQuickReplies = (message?: TMessage): ISuggestion[] =>
...
+  const arrangedMessages = messages.map((message) => {
+    return {
+      ...message,
+      direction:
+        message.author === profile?.foreign_id || message.author === profile?.id
+          ? Direction.sent
+          : Direction.received,
+      read: message.direction === Direction.sent || message.read,
+      delivery: message.direction === Direction.sent || message.delivery,
+    } as TMessage;
```
medchedli: Minor refactor suggestion to help avoid confusing: When we write message.direction in the read and delivery assignments, we're referring to the old message's direction property not the newly computed direction.

```suggestion
    const direction =
      message.author === profile?.foreign_id || message.author === profile?.id
        ? Direction.sent
        : Direction.received;

    return {
      ...message,
      direction,
      read: direction === Direction.sent || message.read,
      delivery: direction === Direction.sent || message.delivery,
    } as TMessage;
```

api/src/extensions/channels/web/base-web-channel.ts
```diff
@@ -131,9 +131,16 @@ export default abstract class BaseWebChannelHandler<
       try {
```
medchedli, line 141: If menu fetching fails, it sends only the basic settings object to the client. The response will not include profile/messages. I suggest we also add profile/messages in the response even when menu fetching has failed.

api/src/extensions/channels/web/base-web-channel.ts
MohamedAliBouhaouala: It's `undefined` by default

widget/src/components/Launcher.tsx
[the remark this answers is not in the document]
MohamedAliBouhaouala: Linting will fail unless we explicitly ignore the dedicated rule `react-hooks/exhaustive-deps`

widget/src/providers/ChatProvider.tsx
```diff
@@ -399,37 +415,44 @@ const ChatProvider: React.FC<{
...
+      });
+    }
+
+    const { quickReplies, arrangedMessages, participantsList } =
+      preprocessMessages(messages, participants, profile);
+
+    setSuggestions(quickReplies);
+    setMessages(arrangedMessages);
+    setParticipants(participantsList);
   });
```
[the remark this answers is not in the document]
medchedli: Is it possible to remove sending GET_STARTED logic from UserSubscription and rely only on sending it when a profile has no messages ?

widget/src/components/Launcher.tsx
```diff
@@ -43,6 +43,14 @@ const Launcher: React.FC<LauncherProps> = ({
 
   useSocketLifecycle();
 
+  useEffect(() => {
+    if (chat.messages.length > 0) {
+      widget.setScreen("chat");
+      chat.setConnectionState(3);
```
marrouchi, line 49:
```suggestion
      chat.setConnectionState(3);
```
Please let's use the enum value instead of "3".

Also, I thought that we update the connection state, from which we infer which screen to display. Or is this done in the subsequent PR ?

widget/src/providers/ChatProvider.tsx
```diff
@@ -46,6 +48,42 @@ export const getQuickReplies = (message?: TMessage): ISuggestion[] =>
           } as ISuggestion),
       )
     : [];
+export const preprocessMessages = (
+  messages: TMessage[],
```
marrouchi, line 52: If `messages` is "undefined" does this break ?

marrouchi (changes requested):

widget/src/providers/ChatProvider.tsx
MohamedAliBouhaouala: `messages` can't be undefined. Worst case scenario it defaults to an empty array.

widget/src/components/Launcher.tsx
MohamedAliBouhaouala: addressed in PR  [1319](https://github.com/Hexastack/Hexabot/pull/1319/)

medchedli (approved):

marrouchi (approved):

PR #1314, opened 2025-08-11
Prevent deletion of blocks configured as Global Fallback block and in use by active conversations

medchedli:
Implement checks to prevent the deletion of blocks that are currently referenced by active conversations. Enhance error handling by displaying toast notifications when block deletion fails.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleting a block is now prevented when it’s used by an active conversation or configured as the global fallback, with clear conflict messages.
  * Visual Editor (bulk delete): errors are now displayed via toast when a delete fails; successful bulk deletes continue to clear the selected block.

### DIFF
--- a/api/src/chat/controllers/block.controller.spec.ts
+++ b/api/src/chat/controllers/block.controller.spec.ts
@@ -9,6 +9,7 @@
 import { ConflictException, NotFoundException } from '@nestjs/common';
 
 import { I18nService } from '@/i18n/services/i18n.service';
+import { SettingService } from '@/setting/services/setting.service';
 import { IGNORED_TEST_FIELDS } from '@/utils/test/constants';
 import { getUpdateOneError } from '@/utils/test/errors/messages';
 import {
@@ -36,6 +37,7 @@ describe('BlockController', () => {
   let blockService: BlockService;
   let categoryService: CategoryService;
   let conversationRepository: ConversationRepository;
+  let settingService: SettingService;
   let category: Category;
   let block: Block;
   let blockToDelete: Block;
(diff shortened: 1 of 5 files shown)


api/src/chat/repositories/block.repository.ts
```diff
@@ -293,6 +297,20 @@ export class BlockRepository extends BaseRepository<
     const docsToDelete = await this.model.find(criteria);
     const idsToDelete = docsToDelete.map(({ id }) => id);
     if (idsToDelete.length > 0) {
+      // Check if any active conversation references this block in current or next
+      const inUse = await this.conversationRepository.model.exists({
+        active: true,
+        $or: [
+          { current: { $in: idsToDelete } },
```
marrouchi, line 306: "current" is not an array of ids, so no need to use the $in operator I think.

api/src/chat/controllers/block.controller.spec.ts
```diff
@@ -78,6 +90,10 @@ describe('BlockController', () => {
     }))!;
   });
 
+  beforeEach(() => {
+    (conversationRepository.model.exists as jest.Mock).mockResolvedValue(false);
+  });
```
marrouchi: Not sure about the utility of this line.

medchedli: This is to make sure that mocked values are always reset to false before running any of the tests, even if a test fails or throws.

api/src/chat/repositories/block.repository.ts
medchedli: Yes `current` is a single ObjectId, but we are checking against multiple block IDs (deleteMany). Using `$in` is still valid and matches if current equals any ID in the list.

frontend/src/components/visual-editor/v2/Diagrams.tsx
```diff
@@ -126,6 +126,19 @@ const Diagrams = () => {
...
+      // Map error codes to i18n keys
+      const errorMap: Record<string, string> = {
+        BlockInUse: "message.block_delete_in_use_conflict",
+        BlockIsGlobalFallback: "message.block_delete_global_fallback_conflict",
+      };
+      const i18nKey = errorType && errorMap[errorType];
+      const localizedFeedbackMessage = i18nKey ? t(i18nKey) : apiErrorMessage;
+
+      toast.error(localizedFeedbackMessage);
+    },
```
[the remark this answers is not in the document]
medchedli: This part of the code was removed from this PR

PR #1323, opened 2025-08-19
Frontend: Add support for block search within the visual editor

### LINKED ISSUE #414: 💡 [REQUEST] - Add Search for Block Functionality in Visual Editor
### Summary

Introduce a search functionality in the visual editor to help users locate specific blocks. Enhance this feature by implementing a zoom effect that focuses on the corresponding block when the user hovers over search results.

Ideally a "autocomplete" alike search that allows to search in title or message object. On click of one of the search result it will navigate to that given block (selects the right flow and re-center and zoom)

medchedli:
# Motivation

Implement full-text search for blocks in the API and integrate search support within the visual editor. This enhances user experience by allowing efficient block searches.

Fixes #414

# Type of change:

# Checklist:

## Summary by CodeRabbit

- **New Features**
  - Block Search panel in the Visual Editor with scope toggle (Current flow / All flows), Cmd/Ctrl+F shortcut, keyboard navigation, result counts, excerpts/icons, and “Show more”/Retry actions.
  - Open/focus a specific block via URL so search selections center the editor view.
  - English and French translations for search, flow, loading, and no-results messages.

- **Style**
  - Editor and canvas layout adjusted to better fill available space; refined node selection visuals and smoother transitions.


frontend/src/utils/block.utils.ts
```diff
@@ -0,0 +1,54 @@
...
+import SimpleTextIcon from "@/app-components/svg/toolbar/SimpleTextIcon";
+import { BlockMessage, BlockType } from "@/types/block.types";
+
+/**
+ * Determines the type of a block message by inspecting its structure and properties.
+ *
+ * @param message - The block message to determine the type of.
+ * @returns The determined `BlockType` for the given message.
+ */
+export const determineType = (message: BlockMessage): BlockType => {
```
marrouchi:
```suggestion
export const getBlockType = (message: BlockMessage): BlockType => {
```

frontend/src/components/visual-editor/hooks/useVisualEditor.tsx
```diff
@@ -266,6 +290,86 @@ const VisualEditorProvider: React.FC<VisualEditorContextProps> = ({
...
+            if (el) {
+              el.classList.add("flash-highlight");
+              if (highlightTimeoutRef.current) {
+                clearTimeout(highlightTimeoutRef.current);
+              }
+              highlightTimeoutRef.current = setTimeout(() => {
+                el.classList.remove("flash-highlight");
+                highlightTimeoutRef.current = null;
+              }, FOCUS_CONFIG.HIGHLIGHT_DURATION_MS);
+            }
```
marrouchi: Ideally we avoid using DOM to update the class. To achieve this, maybe we can re-use the selected block state ?

frontend/src/app-components/displays/ErrorState.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * />
+ *
+ * @example
+ * // Without CTA button
+ * <ErrorState
+ *   message="Permission denied"
+ *   showCta={false}
+ * />
+ */
+export const ErrorState: React.FC<ErrorStateProps> = ({
```
marrouchi: The component name can be misleading. On the other hand, I think MUI has already component to show an error message. Can we re-use that ?

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,485 @@
...
+
+    // Refetch unified data
+    refetchSearch();
+    refetchCategories();
+  };
+
+  if (!open) return null;
+
+  return (
+    <Panel role="dialog" aria-label={t("label.search_blocks_panel_header")}>
```
yassinedorbozgithub:
```suggestion
    <Panel
      role="dialog"
      aria-label={t("label.search_blocks_panel_header")}
      overflow="auto"
    >
```

Before the update (There is no vertical scroll)
image

After the update
image

frontend/src/app-components/displays/ErrorState.tsx
medchedli: MUI has either Alert or Snackbar component to display errors, not sure if they have something else that is suitable for this use case. What if we rename it to ErrorIndicator or ErrorIndicatorView.

frontend/src/utils/block.utils.ts
medchedli: Done ✅

frontend/src/components/visual-editor/BlockSearchPanel.tsx
medchedli: Done ✅

frontend/src/components/visual-editor/hooks/useVisualEditor.tsx
medchedli: Done ✅

frontend/src/app-components/displays/ErrorState.tsx
medchedli: Replaced ErrorState with Snackbar.

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,488 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [error]);
+
+  // Perform search when query changes
+  useEffect(() => {
+    debouncedSearch(visibleSearchItems);
+
+    return () => debouncedSearch.clear();
+  }, [debouncedSearch, visibleSearchItems]);
```
yassinedorbozgithub: This logic is causing an infinite loop and can cause performance issues.

yassinedorbozgithub (changes requested):

frontend/src/components/visual-editor/BlockSearchPanel.tsx
medchedli: Updated the logic to use Use React's useDeferredValue to defer rendering of search results while the user types. This prevents a render loop caused by constantly updating state with new array. Let me know if you have better suggestion to fix the issue.

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,432 @@
...
+    },
+  );
+  // Create a map of category labels by ID for quick lookup
+  const categoryLabelById = useMemo(() => {
+    const map = new Map<string, string>();
+
+    categories.forEach((c) => map.set(c.id, c.label));
+
+    return map;
+  }, [categories]);
```
marrouchi: We can use "useGetFromCache()" instead of constructing a map.

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,432 @@
...
+  // Loading and error states
+  const loading = isLoadingSearch || isLoadingCategories;
+  // Map backend results into UI items
+  const items: BlockSearchItem[] = useMemo(() => {
+    return backendResults.map((blockEntry: IBlockSearchResult) => {
+      const type = getBlockType(blockEntry.message);
```
marrouchi: Note that we can compute the "type" attribute at the processStrategy() level. That way you can update the typescript definition and use "block.type" as if the API returned it. That could be helpful as well in other parts of the code.

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,432 @@
...
+        blockTextContent:
+          typeof blockEntry.message === "string"
+            ? blockEntry.message
+            : Array.isArray(blockEntry.message)
+            ? blockEntry.message.join(" ")
+            : typeof blockEntry.message === "object" &&
+              blockEntry.message &&
+              "text" in blockEntry.message
+            ? String(blockEntry.message.text ?? "")
+            : "",
```
marrouchi: I suggest to declare a helper function "getBlockExcerpt()" that would return such. I wonder though what the content would be when it's a plugin ..

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,432 @@
...
+  }, [backendResults, categoryLabelById]);
+  const visibleSearchItems = useMemo(() => {
+    if (!searchText) return [];
+
+    return items.map((item, idx) => ({
+      item,
+      refIndex: idx,
+    }));
+  }, [searchText, items]);
+  const visibleCount = Math.min(visibleSearchItems.length, shownCount);
```
marrouchi: I suggest honestly to use virtualized list instead and display everything : https://mui.com/material-ui/react-list/#virtualized-list

This will avoid us the complexity of dealing with client side pagination. We can put the limit to 100 (documentation says that it can handle 200 and more).

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,432 @@
...
+            justifyContent="center"
+            sx={{ height: "100%" }}
+            flex={1}
+            gap={2}
+          >
+            <SearchOffIcon sx={{ fontSize: 48 }} />
+            <Typography align="center">
+              {t("message.no_matching_results")}
+            </Typography>
+          </Box>
```
marrouchi: I suggest we use the same component that we use elsewhere to indicate that there's no result. Maybe you could customize the message even more once the related PR is merged (#1323)

frontend/src/components/visual-editor/BlockSearchPanel.tsx
```diff
@@ -0,0 +1,432 @@
...
+            </ResultCount>
+            <Box
+              sx={{
+                flex: 1,
+                minHeight: 0,
+                overflow: "auto",
+              }}
+            >
+              <Box>
+                {visibleSearchItems.slice(0, visibleCount).map((res, idx) => {
```
marrouchi: I suggest we add "BlockSearchResultItem" component in order to reduce complexity and put everything in this component. Let me know what you think.

frontend/src/components/visual-editor/index.tsx
```diff
@@ -34,6 +39,8 @@ export const VisualEditor = () => {
             sx={{
               display: "flex",
               flexDirection: "column",
+              height: "100%",
+              minHeight: 0,
```
marrouchi: Not sure why we need "minHeight: 0," isn't it the default ?

frontend/src/components/visual-editor/v2/CustomCanvasWidget.tsx
```diff
@@ -91,7 +93,11 @@ export class CustomCanvasWidget extends React.Component<DiagramProps> {
     const model = engine.getModel();
 
     return (
-      <div className="canvas-container" tabIndex={0}>
+      <div
+        className="canvas-container"
+        tabIndex={0}
+        style={{ width: "100%", height: "100%", position: "relative" }}
```
marrouchi: By default a div 's width is 100% and the position is relative. Also, maybe better to put everything the css class "canvas-container"

frontend/src/components/visual-editor/index.tsx
medchedli: Yes you are right `minHeight` is not needed here.

frontend/src/components/visual-editor/BlockSearchPanel.tsx
medchedli: Done

frontend/src/components/visual-editor/BlockSearchPanel.tsx
medchedli: Ok. The plugin is a little bit tricky because the data structure is not standard. But will I will try improve the support for plugin blocks

frontend/package.json
```diff
@@ -28,6 +28,7 @@
     "@mui/x-data-grid": "^7.3.2",
     "@projectstorm/react-canvas-core": "^7.0.3",
     "@projectstorm/react-diagrams": "^7.0.4",
+    "@types/react-window": "^1.8.5",
```
marrouchi: This package should be added to the "devDependencies"

frontend/src/components/visual-editor/BlockSearchResultItem.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const getCategoryFromCache = useGetFromCache(EntityType.CATEGORY);
+  const categoryLabel =
+    getCategoryFromCache(String(item.category))?.label ?? "";
+  // Get Block name and excerpt
+  const blockEntryName = item?.name;
+  const blockEntryTextContent = getBlockExcerpt(item.message, item.options);
+  const Icon = getIconForType(item?.type);
+
+  return (
+    <div>
```
marrouchi:
```suggestion
    <div>
```
Can we remove the div ?

frontend/src/components/visual-editor/BlockSearchResultItem.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onClick,
+}) => {
+  // Get category from cache for quick lookup
+  const getCategoryFromCache = useGetFromCache(EntityType.CATEGORY);
+  const categoryLabel =
+    getCategoryFromCache(String(item.category))?.label ?? "";
+  // Get Block name and excerpt
+  const blockEntryName = item?.name;
+  const blockEntryTextContent = getBlockExcerpt(item.message, item.options);
+  const Icon = getIconForType(item?.type);
```
marrouchi:
```suggestion
  const Icon = getBlockIconByType(item?.type);
```

frontend/src/components/visual-editor/hooks/useVisualEditor.tsx
```diff
@@ -33,6 +34,8 @@ import { BLOCK_HEIGHT, BLOCK_WIDTH } from "../v2/CustomDiagramNodes/NodeWidget";
 const engine = createEngine({ registerDefaultDeleteItemsAction: false });
 let model: DiagramModel;
 
+// Focus behavior tuning
```
marrouchi: Is this comment still relevant ?

frontend/src/utils/block.utils.ts
```diff
@@ -0,0 +1,159 @@
...
+ * @returns The determined `BlockType` for the given message.
+ */
+export const getBlockType = (message: BlockMessage): BlockType => {
+  if (typeof message === "string" || Array.isArray(message))
+    return BlockType.TEXT;
+  if (message && typeof message === "object") {
+    if ("attachment" in message) return BlockType.ATTACHMENT;
+    if ("quickReplies" in message) return BlockType.QUICK_REPLIES;
+    if ("buttons" in message) return BlockType.BUTTONS;
+    if ("elements" in message) return BlockType.LIST;
```
MohamedAliBouhaouala, line 34: how about using `switch.. case`  or using a map

frontend/src/components/visual-editor/hooks/useVisualEditor.tsx
```diff
@@ -206,6 +211,30 @@ const buildDiagram = ({
...
+            engine.zoomToFitSelectedNodes({
+              maxZoom: 1,
+              margin: 0,
+            });
+          } catch (_) {
+            // ignore
+          }
+        }, 100);
+      }
+    }
```
MohamedAliBouhaouala, line 237: I don't fully grasp why timeout is being used here ? and are there any alternatives to avoid or mitigate it ?

frontend/public/locales/en/translation.json
```diff
@@ -126,7 +126,9 @@
     "video_error": "Video not found",
     "missing_fields_error": "Please make sure that all required fields are filled",
     "weight_required_error": "Weight is required or invalid",
-    "weight_positive_number_error": "Weight must be a strictly positive number"
+    "weight_positive_number_error": "Weight must be a strictly positive number",
+    "failed_to_load_blocks": "Failed to load blocks search results.",
+    "no_matching_results": "No matching results found. Try other keywords."
```
MohamedAliBouhaouala: `no_matching_results` is unused

frontend/src/components/visual-editor/hooks/useVisualEditor.tsx
medchedli: A timeout is needed to account for a slight delay while the block renders, before applying zoomToFitSelectedNodes.

frontend/src/utils/block.utils.ts
medchedli: I will update it to use Object/Map lookup

frontend/public/locales/en/translation.json
medchedli: I'll update the noDataOverlay component to use this translation key, as it now accepts custom messages.

frontend/src/utils/block.utils.ts
medchedli: On second thought, I believe the current version is better. It's more direct and easier to understand.

MohamedAliBouhaouala: Okay! so the expected behavior is to catch the first type and process it accordingly, right ?

yassinedorbozgithub (changes requested):
Have a minor UI observation
<video src="https://github.com/user-attachments/assets/69f2dbd3-1a9d-46a3-a168-0980903c9a0b"></video>

The selecting height is changing between empty screen and items screen

frontend/src/utils/block.utils.ts
medchedli: Yes exactly.

marrouchi:
Closed in favor of https://github.com/Hexastack/Hexabot/pull/1365

PR #1341, opened 2025-08-29
fix(api): add v2.3.2 options.typing migration

### LINKED ISSUE #1340: 🐛 [BUG] - migration file to make options.typing number
### Description

The zod validation will check if option.typing have a number type.
We will create a migration file to transform all option.typing to have a number type.

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

```bash

```

### Browsers

_No response_

### OS

_No response_

yassinedorbozgithub:
# Motivation
The main motivation of this PR is to add a migration to transform the block options.typing to be a number.

Fixes #1340

# Type of change:

# Checklist:

## Summary by CodeRabbit

- **Chores**
  - Performed a data migration to standardize the “typing” setting in blocks to a consistent numeric format, improving data consistency and stability across the app.
  - Migration runs automatically during upgrade; no user action required.


api/src/migration/migrations/1756483008605-v-2-2-12.migration.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  const adminUser = await getAdminUser();
+
+  if (!adminUser) {
+    services.logger.warn('Unable to process block, no admin user found');
+    return;
+  }
+
+  try {
+    await BlockModel.updateMany({ 'options.typing': { $exists: true } }, [
```
MohamedAliBouhaouala: Quick question: what if a single documents contains an invalid `options.typing` field and thus cannot be parsed to int ?

yassinedorbozgithub: This is a good edge case, but we are not supposed to face this case except if the data is updated not from the UI.
This migration main motivation is to resolve **block.options.typing** values previously set it accidentally with a string values (UI is fixed to not introduce new values with not expected type).

PR #1356, opened 2025-09-03
fix: ws connection mutex

marrouchi:
# Motivation
This PR introduces a safer, cross-tab–aware reconnect() for our Socket.IO client.
It uses the Web Locks API to serialize reconnection attempts across tabs, adds timeouts and error handling, and ensures listener cleanup so the lock is never held indefinitely. It also switches state checks to socket.connected for correctness.

## Problem

When multiple tabs try to reconnect at the same time, we can:
- spin up parallel WS handshakes and duplicate server-side session work,
- deadlock if connect never fires (no timeout / no reject),
- leak listeners, keeping the lock until reload,
- rely on socket.active, which is not a reliable indicator of connection state.

## Solution
- Wrap the reconnect flow in navigator.locks.request('ws_connection', { mode: 'exclusive' }, …) to serialize attempts across tabs.
- Use a promise with timeout (default: 10s) and listen to connect, connect_error, and error to resolve/reject deterministically.
- Always cleanup (remove listeners + clear timeout) in a finally block to guarantee lock release.
- Prefer socket.connected to short-circuit when already connected and to decide if we should force a fresh handshake.
- Graceful fallback: if Web Locks isn’t available, run the reconnect logic without a lock (behavior matches current, but with safer timeout/error handling).

## Other updates
- Prevent from subscribing multiple times to broadcast channel when we have multiple renders
- Prevent "get started" from being sent multiple times

# Type of change:

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Cross-tab coordinated reconnect so the widget recovers reliably across browser tabs.
  * Added an unsubscribe API for broadcast subscriptions.

* **Bug Fixes**
  * Added unsubscribe/cleanup to prevent duplicate broadcast subscriptions and stray handlers.
  * Improved handling of socket/auth interruptions to reduce reconnection stalls.
  * Removed a redundant cross-tab broadcast call in one component.

* **Refactor**
  * Unified reconnection flow to use a single reconnect action across UI flows.

* **Chores**
  * Updated copyright year to 2025.

### DIFF
--- a/widget/src/providers/ChatProvider.tsx
+++ b/widget/src/providers/ChatProvider.tsx
@@ -460,8 +460,7 @@ const ChatProvider: React.FC<{
   });
 
   useSubscribeBroadcastChannel("subscribed", () => {
-    socket.disconnect();
-    socket.connect();
+    socket.forceReconnect();
   });
 
   useSubscribe("error", ({ message, statusCode }: SocketErrorResponse) => {
@@ -473,9 +472,7 @@ const ChatProvider: React.FC<{
   useSubscribe("settings", ({ profile, messages = [] }: ChannelSettings) => {
     setProfile(profile);
 
-    if (config.channel === "web-channel" && profile && messages.length === 0) {
-      sendGetStarted(profile.foreign_id);
-    } else if (config.channel === "console-channel" && !profile) {
+    if (config.channel === "console-channel" && !profile) {
       handleSubscription();
     }
 
--- a/widget/src/providers/ChatProvider.tsx
+++ b/widget/src/providers/ChatProvider.tsx
@@ -40,6 +40,7 @@ import {
 } from "../types/state.types";
 import { SocketIoClientError } from "../utils/SocketIoClientError";
 
+import { useBroadcastChannel } from "./BroadcastChannelProvider";
 import { useConfig } from "./ConfigProvider";
 import { ChannelSettings, useSettings } from "./SettingsProvider";
 import { useSocket, useSubscribe } from "./SocketProvider";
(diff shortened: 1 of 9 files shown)


widget/src/utils/SocketIoClient.ts
```diff
@@ -118,6 +118,84 @@ export class SocketIoClient {
     if (!this.socket.active) this.socket.connect();
   }
 
+  /**
+   * Waits for disconnecttion of the socket client.
```
medchedli:
```suggestion
   * Waits for disconnection of the socket client.
```

widget/src/providers/ChatProvider.tsx
```diff
@@ -434,6 +436,11 @@ const ChatProvider: React.FC<{
         author: foreign_id,
       },
     });
+
+    // Notify other tabs that we've subscribed and sent the 1st msg (so they reload)
+    setTimeout(() => {
+      postMessage({ event: "subscribed" });
+    }, 500);
```
MohamedAliBouhaouala, line 443: Let's add a `TODO` comment here so that we can focus on finding better alternatives later

marrouchi: Yup, let's open an issue instead. @coderabbitai Please open a new issue for the chat widget about "postMessage({ event: "subscribed" });"  It allows to broadcast the fact that user has subscribed (submitted the form) to start the chat. It uses the Broadcast Channel API in order to notify the other tabs of this state change. We have a racing condition where the other tabs could reload before the chatbot responds with the initial "Get Started" message. Currently we use timeout in order to address this but we need to find a better way.

MohamedAliBouhaouala (approved):

PR #1461, opened 2025-11-12
fix(frontend): resolve hotreload detection logic

yassinedorbozgithub:
# Motivation
Resolve hotreload detection logic

# Type of change:


packages/frontend/src/hooks/useApiClient.ts
```diff
@@ -59,14 +57,14 @@ export const useAxiosInstance = () => {
 
         if (
           error.response.status === 500 &&
-          error.request.responseURL.endsWith(ME_ENDPOINT)
+          error.request.responseURL.endsWith(`/api${ROUTES.ME}`)
```
marrouchi, line 59: I think we need to add the "api" prefix only when monolith mode is enabled.

PR #1488, opened 2025-11-18
feat: create a custom template mjmlAdapter

yassinedorbozgithub:
# Motivation
create a custom template mjmlAdapter

# Type of change:

# Checklist:

### DIFF
--- a/packages/api/src/mailer/mjml-adapter.class.ts
+++ b/packages/api/src/mailer/mjml-adapter.class.ts
@@ -19,7 +19,6 @@ export class MjmlAdapter implements TemplateAdapter {
   ) => {
     try {
       const templatePath = path.join(
-        process.cwd(),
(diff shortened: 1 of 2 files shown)


packages/api/src/mailer/patched-mjml-adapter.class.ts
```diff
@@ -0,0 +1,37 @@
...
+import Handlebars from 'handlebars';
+
+export class FixedMjmlAdapter implements TemplateAdapter {
+  compile: TemplateAdapter['compile'] = async (
+    mail: { data: ISendMailOptions },
+    callback,
+    options,
+  ) => {
+    try {
+      const path = `${options.template?.dir}/${mail.data.template}`;
```
marrouchi: path.join() is better I think also let's remove patch-package npm dep.

PR #1576, opened 2026-01-14
fix(frontend): add memory definitions to workflow types

yassinedorbozgithub:
# Motivation
Add memory definitions to workflow types


packages/frontend/src/types/workfow.types.ts
```diff
@@ -24,6 +24,7 @@ export interface IWorkflowAttributes {
   x?: number;
   y?: number;
   direction?: ResizeControlDirection;
+  memoryDefinitions: [];
```
marrouchi:
```suggestion
  memoryDefinitions: string[];
```

marrouchi (changes requested):

PR #1745, opened 2026-02-23
feat: add credentials management

yassinedorbozgithub:
# Motivation
Add credentials management


packages/api/src/user/repositories/credential.repository.ts
```diff
@@ -0,0 +1,41 @@
...
+    super(repository);
+  }
+
+  async findOneByIdWithPassword(
+    id: string,
+  ): Promise<CredentialOrmEntity | null> {
+    return await this.repository.findOne({
+      where: { id },
+    });
+  }
```
marrouchi: Let's keep using findOne() which does the same.

packages/api/src/extensions/actions/llm/agent.action.ts
```diff
@@ -54,9 +54,14 @@ export class LlmAgentAction extends LlmBaseAction<
     const logger = context.services.logger;
     const providerName = settings.provider ?? 'openai';
     const modelId = this.resolveModelId(settings);
+    const credentialRepository =
+      await context.services.credential.repository.findOneByIdWithPassword(
```
marrouchi:
```suggestion
      await context.services.credentials.findOne(
```

packages/api/src/workflow/contexts/workflow-runtime.context.ts
```diff
@@ -66,6 +70,7 @@ export abstract class WorkflowRuntimeContext<
       contentType: this.contentType,
       message: this.message,
       actions: this.actionService,
+      credential: this.credentialService,
```
marrouchi:
```suggestion
      credentials: this.credentialService,
```

marrouchi (changes requested):

PR #1749, opened 2026-02-23
fix(cli): remove CLI help when running the Project

yassinedorbozgithub:
# Motivation
Remove CLI help when running the Project

# Before the fix
> pnpm --filter @hexabot-ai/cli dev
image

> pnpm dev
image

> pnpm --filter @hexabot-ai/api cli
image

# After the fix
> pnpm --filter @hexabot-ai/cli dev
image

> pnpm dev
image

> pnpm --filter @hexabot-ai/api cli
image


packages/cli/src/index.ts
```diff
@@ -9,13 +9,21 @@ import { createCliProgram } from './cli.js';
 import { checkPrerequisites } from './core/prerequisites.js';
 import { printBanner } from './ui/banner.js';
 
+const cliArgs = process.argv.slice(2);
+const suppressIdleOutput =
+  process.env.HEXABOT_CLI_DEV_IDLE === '1' && cliArgs.length === 0;
```
marrouchi: Can we avoid introducing a new env var ?

marrouchi (changes requested):

packages/cli/src/index.ts
yassinedorbozgithub: Yes, i will do that :technologist:

PR #1756, opened 2026-02-24
fix(api): resolve updating user roles event emitting

yassinedorbozgithub:
# Motivation
Resolve updating user roles event emitting


packages/api/src/user/repositories/user.repository.ts
```diff
@@ -25,6 +29,24 @@ export class UserRepository extends BaseOrmRepository<
...
+      ['many-to-many'],
+      payload,
+    );
+
+    if (hasOnlyRelationTypeProperties) {
+      await this.emitHookOnUpdateRelations(idOrOptions, updated);
+    }
+
+    return updated;
+  }
```
marrouchi: I would suggest we find a solution that adresses all uses cases not only for the user relations.

marrouchi (changes requested):

packages/api/src/user/repositories/user.repository.ts
yassinedorbozgithub: Happy to do that :technologist:

packages/api/src/utils/generics/base-orm.repository.ts
```diff
@@ -284,6 +284,20 @@ export abstract class BaseOrmRepository<
...
+      const hasOnlyManyToManyProperties =
+        entries.length > 0 &&
+        entries.every(
+          ([key, val]) => manyToManyKeys.includes(key) && Array.isArray(val),
+        );
+
+      if (hasOnlyManyToManyProperties) {
+        entity.updatedAt = new Date();
+      }
+
```
marrouchi, line 304: Can you please add a comment the explain the motivation behind this ?

marrouchi (changes requested):

packages/api/src/utils/generics/base-orm.repository.ts
yassinedorbozgithub: Sure, happy to do that :technologist:

PR #1786, opened 2026-03-16
fix(frontend): resolve WS entities synchronization issues

yassinedorbozgithub:
# Motivation
Resolve WS entities synchronization issues

### DIFF
--- a/packages/frontend/src/services/entities.ts
+++ b/packages/frontend/src/services/entities.ts
@@ -11,6 +11,16 @@ import { ISubscriberStub } from "@/types/subscriber.types";
 
 import { EntityType } from "./types";
 
+export const applySubscriberDerivedFields = <T extends ISubscriberStub>(
+  subscriber: T,
+): T => {
+  if (subscriber.firstName && subscriber.lastName) {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/services/entities.ts
```diff
@@ -63,9 +73,7 @@ export const SubscriberEntity = new schema.Entity(
         entity.retainedFrom = new Date(entity.retainedFrom);
       }
 
-      if (entity.firstName && entity.lastName) {
-        entity.fullName = `${entity.firstName} ${entity.lastName}`;
-      }
+      entity.fullName = applySubscriberDerivedFields(entity).fullName;
```
marrouchi, line 76: If we add fullName, shouldn't we update the typing as well ? That said, maybe this should go to another PR cause we need refactor many parts of the code ?

Also I think we can apply this not only to subscriber but also to the user.

marrouchi (changes requested):

packages/frontend/src/services/entities.ts
yassinedorbozgithub: Happy to apply the changes :technologist:

PR #1790, opened 2026-03-16
fix(frontend): update subscriber types

yassinedorbozgithub:
# Motivation
Update subscriber types


packages/frontend/src/layout/Header.tsx
```diff
@@ -141,7 +143,7 @@ export const Header: FC<HeaderProps> = ({ isSideBarOpen, onToggleSidebar }) => {
                   lineHeight={1}
                   textTransform="capitalize"
                 >
-                  {user?.firstName} {user?.lastName}
+                  {userFullName || user?.email}
```
marrouchi:
```suggestion
                  {user.fullName || user?.email}
```
We can use fullName directly.

marrouchi (changes requested):

PR #1930, opened 2026-04-27
chore(frontend): remove dead translation keys

yassinedorbozgithub:
## What changed?
This PR removes dead i18n translation keys

## Summary by CodeRabbit

* **Chores**
  * Removed a large set of English and French UI translation entries; core labels remain.
  * Result: some conversational trigger, message/meta, schema, and workflow labels were removed and may fall back to defaults or appear untranslated in the UI.


yassinedorbozgithub:
@coderabbitai, double check if any of the removed translation keys still used and need to be recovered

PR #1931, opened 2026-04-27
fix(api): block prototype pollution in TypeORM search filter pipe

anis-marrouchi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened path validation to block unsafe nested property names, preventing prototype-pollution and ensuring invalid sort fields fall back to the default ordering.
* **Tests**
  * Added parameterized tests validating path-safety across parsing, sorting, and object-merge scenarios to ensure robust fallback behavior and protection against injection attempts.


marrouchi:
Thanks @noqta_coder_bot, can we apply FORBIDDEN_KEYS check in parseSort method as well to prevent prototype pollution via untrusted sort field.

Also, maybe better to scan other other pipes to check if we need to apply the same safe-guard elsewhere and centralize the FORBIDDEN_KEYS check.

anis-marrouchi:
Closing per maintainer request — reverting coder agent changes.

PR #1955, opened 2026-04-29
fix(api): holistic prototype pollution guard in base-orm repository

anis-marrouchi:
## Summary
- Adds `assertSafePayload()` validation to `create`, `createMany`, `updateOne`, and `updateMany` in `BaseOrmRepository`
- Reuses the existing `hasForbiddenSegment` helper to reject `__proto__`, `constructor`, and `prototype` keys in flattened payloads before they reach the ORM layer
- Extends the fix from #1940 (search filter) to cover all write paths

## Context
This is a follow-up to #1940 which addressed prototype pollution in search filters. While that PR hardened the query/filter path, the create and update paths in `BaseOrmRepository` were still unguarded. This PR closes that gap.

Closes #5

## Test plan
- [x] 7 new test cases covering `__proto__`, `constructor`, `prototype` in create/createMany/updateOne/updateMany
- [x] Safe payload passthrough test to avoid false positives
- [x] Pre-commit hooks pass with related tests only

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Write operations (create, createMany, updateOne, updateMany) now enforce payload safety and reject requests containing forbidden property patterns, including nested keys and array-contained values.

* **Tests**
  * Added tests verifying rejection of unsafe payloads (various nesting/array scenarios) and acceptance of valid payloads.

### DIFF
--- a/packages/api/src/utils/generics/base-orm.repository.spec.ts
+++ b/packages/api/src/utils/generics/base-orm.repository.spec.ts
@@ -6,7 +6,7 @@
 
 import { randomUUID } from 'crypto';
 
-import { NotFoundException } from '@nestjs/common';
+import { BadRequestException, NotFoundException } from '@nestjs/common';
 import { EventEmitter2 } from '@nestjs/event-emitter';
 import { TestingModule } from '@nestjs/testing';
 import { In, InsertEvent, RemoveEvent, Repository, UpdateEvent } from 'typeorm';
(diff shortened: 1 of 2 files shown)


marrouchi:
Thank you @anis-marrouchi

If I'm not mistaken, prototype pollution could affect object fields not all the payload. Meaning that only TypeORM entity attributes that are decorated with @JsonColumn() are most likely to be a target of such a vulnerability.

I wonder if we could rather update the actionDtoToEntity() method to look into the metadata of the attribute in order to assess that sanitization is needed or not. I wonder also how TypeORM handles JSON column data, it may be that TypeORM has that protection mechanism already.

PR #2011, opened 2026-05-16
feat(frontend): intial guided tour

marrouchi:
## What changed?

- Add an admin workflow guided tour that helps eligible users create and test their first conversational workflow.
- Automatically starts from the dashboard, creates a starter workflow, and continues through the visual editor.
- Guides users through inserting a step, choosing Send Text Message, saving the action, and testing it in the chat widget.
- Tracks tour completion per user so the tour does not restart after it is finished or skipped.
- Adds improved tooltip UI and friendlier English/French tour copy.

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -59,6 +59,7 @@
     "react-hook-form": "^7.51.5",
     "react-i18next": "^14.1.1",
     "react-is": "^18.3.1",
+    "react-joyride": "^2.9.3",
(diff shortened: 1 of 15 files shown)


yassinedorbozgithub (approved):
I’ve tested the initial tour feature, and it looks excellent 🚀
<video src="https://github.com/user-attachments/assets/0ef41306-eecb-4dae-b226-2d5f354b7416"></video>

PR #2092, opened 2026-07-03
feat: enhance trigger panel UX/UI

marrouchi:
## What changed?

Manual workflows can now be triggered by external services through a public webhook endpoint, POST /webhook/:id/trigger. The webhook is opt-in per workflow, supports several authentication schemes, and the UI helps users configure it and integrate with it via ready-to-copy code snippets.

### API
- New WebhookTriggerGuard protecting the endpoint: resolves the workflow, requires it to be a manual workflow with the webhook enabled (returning 404 otherwise so existence isn't disclosed), and verifies credentials with timing-safe comparisons.
- Four auth modes on the webhookTrigger config: none, Basic, custom header, and JWT (algorithm configurable). Zod schema enforces that credentials are set whenever the webhook is enabled.
- New WebhookTriggerService runs the workflow synchronously with the request body as workflow input and returns the run's final status and output.

### Frontend
- Workflow settings (WorkflowForm): new webhook trigger section to enable the webhook and configure the auth mode and credentials.
- Code snippet dialog: generates copyable integration snippets (cURL, Wget, fetch, Python requests, axios) from the workflow's URL, auth config, and current input payload.
- Trigger simulator panel: shows a "Webhook enabled" chip with quick access to the snippet dialog. Also reworked its layout — the panel now fills its drawer column with the input form scrolling independently, the Run button is a full-width action footer with a loading state, and empty/hint states are proper centered illustrations instead of raw text.
- Extracted useRunWorkflow hook so the run mutation (CSRF, toasts, cache invalidation) is shared between the run header and the simulator panel; manual workflows now run from the simulator next to their input form.


yassinedorbozgithub (commented):
Tests all the authentication methods from Postman ✅ 
Hope only to notice that for JWT the secret need to follow some specifications 

Image

PR #2094, opened 2026-07-04
fix: create workflow anomaly

marrouchi:
## What changed?
After creating a new workflow from the visual editor ("New" → fill form → submit), the editor displayed a completely different workflow graph instead of the new blank one.

Two independent issues stacked up:

1. API — missing/incomplete mutation events on workflow creation. The blank initial version (and the workflow's currentVersion pointer) is persisted by entity lifecycle hooks through the raw EntityManager, which bypasses repository-level mutation events. Clients never received the version's postCreate event, the create response/postCreate event carried a stale currentVersion, and the pointer update on snapshot creation went through the raw TypeORM repository without broadcasting a workflow postUpdate.

2. Frontend — stale-collection race in the editor's fallback redirect. WorkflowProvider treats "flowId not in the cached workflows collection" as "workflow doesn't exist" and bounces to workflows[0] (the oldest workflow). Right after a create or import, the new workflow is legitimately absent from the collection until the websocket-triggered refetch lands, so the guard hijacked the navigation to the new workflow — synchronously, before any refetch could win the race.

## Changes

1. API

- WorkflowService.create() now emits the blank initial version's postCreate event after commit (via new WorkflowVersionRepository.emitPostCreateById()).
- The in-memory workflow entity mirrors the currentVersion pointer set by lifecycle hooks, so create responses, import postCreate events, and transfers carry the final pointer.
- WorkflowVersionService.createSnapshot() re-applies the currentVersion pointer through the service layer so the workflow postUpdate event is broadcast (restore no longer duplicates this).

2. Frontend

The WorkflowProvider fallback redirect now also checks the entity item cache before bouncing: a just-created/imported workflow is present there synchronously, so navigation stays on it while the collection refetch catches up. Deleted workflows still redirect correctly since the WS delete handler removes their item-cache entry. Cross-tab sync remains fully WS-driven — no local cache patching or extra invalidation.

### DIFF
--- a/packages/api/src/transfer/workflow-transfer.service.ts
+++ b/packages/api/src/transfer/workflow-transfer.service.ts
@@ -506,6 +506,11 @@ export class WorkflowTransferService {
           workflowVersionPayload,
         );
 
+      // The version's lifecycle hook moved the workflow's currentVersion
+      // pointer on a separately-loaded row; mirror it on the in-memory entity
+      // so the deferred workflow postCreate event carries the final pointer.
+      importedWorkflow.workflowEntity.currentVersion = workflowVersion;
+
       postCreateEvents.push(
         buildPostCreateEvent(
           'workflowVersion',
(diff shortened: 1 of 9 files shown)


yassinedorbozgithub (approved):
LGTM

marrouchi:
Closed in favor of https://github.com/hexabot-ai/Hexabot/pull/2095

PR #2101, opened 2026-07-06
Enhance schedule UI for scheduled workflows

marrouchi:
## What changed?

Improves the UX of the CronInput component used when configuring scheduled workflows, and relocates it within the workflow form.


yassinedorbozgithub (approved):
LGTM

Image

PR #2116, opened 2026-07-09
fix(graph): enhance graph performance

yassinedorbozgithub:
## Summary
Optimized the workflow layout logic by centralizing shared graph, geometry, and branch traversal helpers while keeping the same layout behavior. Also refined branch alignment to avoid nested group spacing regressions and prevent overlap between unrelated nested branches.

## Why
The layout code had repeated logic across multiple passes, making it harder to maintain and easier to introduce inconsistent behavior. The update reduces duplicated code while preserving visual stability for nested groups, branch placeholders, and branch-origin alignment.

## How to review
Focus on the centralized helpers in `geometry.ts` and `graph-maps.ts`, then review the updated layout passes that consume them. Pay special attention to:
- nested group bounds and spacing
- first-level branch alignment
- branch symmetry and flow-origin alignment
- trailing placeholder positioning
- attachment/subtree movement behavior

## Validation
Checked the visual workflow layout with nested conditionals, grouped branches, placeholders, and trailing continuation nodes. Verified that the first-level branches align correctly without pulling deeper nested branches into the same column or causing overlaps.


marrouchi:
This PR has been closed to avoid introducing more complexity.

PR #2118, opened 2026-07-10
feat(api): make the ai agent step budget a workflow-global setting

### LINKED ISSUE #1977: [Issue] AI agent default stop condition undercounts tool-call budget when MCP tools are bound, preventing memory updates
### Details

AI actions currently derive the default `stopWhen` condition from the number of available tools, using roughly:

```ts
defaultStepCount = 1 + Object.keys(tools).length
```

This default is too restrictive for agent workflows that bind MCP servers and memory. An MCP binding may expose several concrete tools, and a successful task may require multiple calls to the same MCP server before the agent can call other tools such as `update_memory`.

In workflows where the outer flow uses a `while` condition based on memory state, this can cause the agent to stop before writing memory. The loop condition then remains true, so the workflow keeps asking the user / re-running the agent indefinitely.

**Example Scenario**

A scheduling workflow loops until memory contains:

```yaml
$context.memory.appointment_schedule.scheduled
```

The `ai_schedule_agent` has:

```yaml
bindings:
  memory:
    - schedule_memory
  model: model
  mcp:
    - google_calendar
```

The agent may need to call multiple Google Calendar MCP tools, for example:

1. check calendar availability
2. create event
3. generate / attach Google Meet link
4. update memory with `scheduled: true`

With the current default stop behavior, the agent can exhaust its step budget while using MCP tools and never reaches `update_memory`. Since the memory value is not updated, the workflow-level `while` condition stays true and the conversation can loop forever.

**Current Behavior**

- AI actions default the stop step count based on the number of bound/generated tools.
- The default assumes each available tool only needs to be called once.
- MCP-heavy tasks often require several tool calls before completion.
- Memory update can be starved by earlier MCP calls.
- Workflow loops depending on memory state can become non-terminating.

**Expected Behavior**

The default stop condition should give the agent enough room to complete realistic multi-step tool workflows, especially when MCP and memory bindings are present.

At minimum, the default should not prevent the agent from calling `update_memory` after MCP operations.

**Impact**

- Agents using MCP servers may fail to complete tasks.
- Memory-driven workflow loops can become infinite.
- Users must manually guess and configure `stop_step_count` for each MCP workflow.
- The behavior is surprising because adding an MCP server can silently reduce the agent’s ability to reach memory tools.

morgan-coded:
`stop_step_count` was hidden, and its implicit default changed with the number of bound tools, so users could not see or configure a stable limit when a workflow stopped early. The setting is now visible with an advertised default of 10, and an explicit value still wins. Tool-enabled workflows, including those with more than nine tools, now use 10 unless configured higher; tool-less runs and `stop_tool_call` composition keep their existing behavior. Fixes #1977


marrouchi:
@morgan-coded Thank you for the contribution! The solution looks good and addresses the immediate issue.

My main concern is that introducing a hard-coded minimum of 10 creates implicit behavior that is not visible or configurable by the end user. At the very least, this would need to be clearly documented. This lack of visibility is one of the reasons we have not addressed the issue yet, we want to avoid replacing one surprising default with another.

How about considering introducing a workflow-level setting that lets users define the maximum tool-call or step count, defaulting to 10. This would provide a sensible default while making the execution budget explicit and configurable for MCP-heavy workflows.

What do you think about this approach? I see this as both a UX and DX concern: users should not have to understand an internal hard-coded constant when debugging why an agent stopped before completing its task.

morgan-coded:
Thanks, that makes sense. Reworked the step budget as a visible workflow setting with a default of 10; an explicit value still wins, and the tool-count formula plus the MCP-and-memory-only floor are gone. Tool-enabled workflows now default to 10 even when they expose more than nine tools, so larger budgets need to be set explicitly. The focused tests, API checks, and full workspace checks are passing.

packages/api/src/extensions/actions/ai/ai-schemas.ts
```diff
@@ -180,17 +182,12 @@ export const aiCommonSettingsSchema = z.strictObject({
...
-    .optional()
-    .meta({
-      title: 'Stop step count',
-      description: 'Maximum number of agent steps before stopping.',
-      'ui:options': {
-        hideUntilAdded: true,
-      },
-    }),
+  stop_step_count: z.int().positive().optional().meta({
+    title: 'Stop step count',
```
yassinedorbozgithub: Hi @morgan-coded, do think adding a max value .max(MAX_AI_STEP_BUDGET) can maybe help to avoid any accidental runaway costs

yassinedorbozgithub (commented):
Hi @morgan-coded, do you think adding a UI message would help users understand what's happening during the final stage of the execution?

morgan-coded:
Capped `stop_step_count` at 100 in the settings schema, while keeping the default at 10 and accepting configured values through the boundary. I used 10× the default to leave headroom for longer workflows while rejecting accidental four- or five-digit budgets; happy to tune that ceiling if you prefer a different value. Kept the final-stage UI message out of this revision so the cap stays focused, and can follow up with it separately if you'd like.

yassinedorbozgithub:
> Capped `stop_step_count` at 100 in the settings schema, while keeping the default at 10 and accepting configured values through the boundary. I used 10× the default to leave headroom for longer workflows while rejecting accidental four- or five-digit budgets; happy to tune that ceiling if you prefer a different value. Kept the final-stage UI message out of this revision so the cap stays focused, and can follow up with it separately if you'd like.

Hi @morgan-coded, thank you :pray: for being reactive :zap: 
Sure no problem, with follow upping "the reached max step count" UI error message separately in a new PR.

Please don't forget @marrouchi's feedback about making DEFAULT_AI_STEP_BUDGET a workflow global setting
image

morgan-coded:
Implemented @marrouchi's workflow-global default. There's now an `AI agent` settings group with `default_ai_step_budget` set to 10 by default, and the AI actions use it whenever `stop_step_count` isn't set on the action.

The per-action value still takes precedence, and the 100-step cap is unchanged. Because this is a new settings group, bootstrap seeding adds it to existing installs without a migration.

I'm keeping the "reached max step count" UI message in a separate follow-up PR as agreed.

yassinedorbozgithub:
@morgan-coded, i can't see stop_step_count as global setting after your changes :technologist: 
image

morgan-coded:
Placed it as a site-wide `AI agent` settings group, so it lives under the main Settings rather than the workflow's Global Settings dialog — which is why it's missing from your screenshot. From that view you want it beside Timeout/Retries and overridable per step, which makes sense; I can move it there.

Concretely that means adding the step-budget field to the shared workflow settings schema (`BaseSettingsSchema`) so it renders in that dialog, with the AI actions reading the default from the workflow settings instead of a global one. The per-action `stop_step_count` override and the 100 cap would stay as-is.

One thing worth your call: that puts an AI-specific field into the generic workflow settings next to Timeout/Retries. Good to go that route, or would you rather it stay AI-scoped somewhere else?

yassinedorbozgithub:
@morgan-coded, yes move them to `BaseSettingsSchema` double check the global settings UI 
and it will be perfect if you can make the PR changes as minimal as possible :pray:

morgan-coded:
Moved `stop_step_count` into `BaseSettingsSchema`, so it now renders beside Timeout/Retries in the workflow Global Settings dialog and uses the same per-step `Override global settings` path. Removed the separate admin `ai_settings` group, kept the 100-step cap, and rebased the branch onto current `main` with this follow-up limited to the schema move and its tests. Could you confirm the dialog looks right on your side?

yassinedorbozgithub (changes requested):
Hi @morgan-coded, can't see stop_step_count as global setting after your changes 🧑‍💻
image

morgan-coded:
Reproduced the schema-to-form path at `e8da9375` from a fresh checkout. `stop_step_count` is in the workflow title bar's Global Settings dialog beside Timeout/Retries, not the site-wide admin Settings page; the same base schema feeds each step's `Override global settings` panel. After `pnpm install` and `pnpm --filter @hexabot-ai/agentic run build`, both the built and source schemas list `["timeout_ms","retries","stop_step_count"]`, and the workflow form schema exposes the label `Stop step count`. Could you retry after rebuilding `@hexabot-ai/agentic`, restarting the frontend, and opening the workflow title bar's Global Settings dialog?

yassinedorbozgithub:
Hi @morgan-coded, i can see **Stop Step Count** as global setting :tada: 

Can you please (By applying the recommended changes the PR will be ready to be merged)
- Make Stop Step Count before Retries disabled section
- Remove unnecessary file changes (Leave the minimal changes having the expected behavior)

image

Thanks

morgan-coded:
Reordered the shared schema so Stop Step Count now renders before Retries in both workflow settings forms. Trimmed the branch diff from 7 files (+538/-24) to 6 files (+66/-23) while preserving the final behavior and its regression coverage. Local tests are passing. Back to you for another look.

PR #2120, opened 2026-07-12
fix: duplicate child workflow execution on resume

### LINKED ISSUE #2119: call_workflow re-spawns the child workflow when the parent resumes (duplicate child run + orphaned suspension)
**Version:** @hexabot-ai/api 3.3.x, @hexabot-ai/agentic 3.2.1 (conversational workflows, WhatsApp channel)

## Summary

`call_workflow`'s execute performs its side effect (spawning the child run via `agentic.callWorkflow(...)`) **before** calling `context.workflow.suspend(...)`. When the child finishes and `resumeParentRunFromChild` resumes the parent, deterministic replay re-executes the suspended step's action — so the spawn runs a **second time**:

1. A duplicate child run is created; its opening steps re-execute visibly (in our case it re-sent its "send me a photo" ask to the WhatsApp user).
2. The duplicate suspends on its own `await_reply` and is orphaned in `suspended` forever. Because `findSuspendedRunByInitiator` prefers the deepest suspended leaf, the orphan later swallows an unrelated inbound message.

The parent itself completes correctly (its `suspend()` consumes the recorded resume data), which makes the duplicate easy to miss.

## Repro

- Workflow A: `call_workflow` → workflow B.
- Workflow B: `send_text_message` → `await_reply`.
- Trigger A; reply to B's message.
- Observe `workflow_runs`: a second B run is created at the moment the reply arrives (timestamped with the parent resume), stuck `suspended` at the same `await_reply` step, and B's opening message is sent twice.

## Suggested fix

Guard the spawn against replay. The action can detect the replay case cheaply: on a parent resume, `runner.getLastResumeData()` is exactly the `toCallWorkflowResult(...)` payload for the awaited child — if it parses against the resume schema and matches the step's `workflow_id`, skip `callWorkflow(...)` and go straight to `suspend()` (which returns the recorded result). We've deployed a local action with that guard and it resolves both the duplicate sends and the orphaned runs. Alternatively, the engine could cache/replay pre-suspend action side effects, or `resumeParentRunFromChild` could prime the resume so the step short-circuits before executing.

Happy to share more run data if useful. (Related earlier reports from the same deployment: POST /api/translation/refresh self-deadlocks its DB pool beyond ~7 strings; nested `=` expressions inside array/object input literals are recompiled without registered JSONata functions, breaking `$t` — T1006.)

marrouchi:
### Problem

`call_workflow` spawns its child run **before** calling `context.workflow.suspend()`. When the child finishes and the parent is resumed, deterministic replay re-executes the suspended step's action body — so the spawn ran a second time. The result: a duplicate child run whose opening steps re-executed visibly (e.g. its first message sent twice), left orphaned in `suspended` forever, where it would later swallow an unrelated inbound message via `findSuspendedRunByInitiator`. The parent itself completed correctly, which made the duplicate easy to miss.

Issue : #2119 

### Fix (two layers)

**Engine (`@hexabot-ai/agentic`):** new `WorkflowRuntimeControl.hasRecordedResult(key?)` lets an action detect that its upcoming `suspend()` will resolve immediately from a recorded replay result, so side effects placed before the suspend can be skipped. This gives every side-effecting-then-suspending action a correct pattern, not just `call_workflow`.

**API:**
- `call_workflow` now guards its spawn with `hasRecordedResult()` and goes straight to `suspend()` on replay.
- `AgenticService.callWorkflow` adds defense in depth: when the runtime signals a replay *and* the parent's persisted suspension metadata (`awaiting_child_workflow` + linked `workflow_run_id`) matches, it returns the existing child run's persisted result instead of spawning. The runtime signal is required because suspension bookkeeping isn't cleared on resume — metadata alone would misfire on a later sequential call to the same workflow (covered by a dedicated test).

### Tests

- Engine: replay of a rebuilt runner no longer re-runs guarded side effects; `hasRecordedResult` reports per-suspension replay state across multi-suspend steps.
- API: new `call-workflow.action.spec.ts`; service-level tests assert no duplicate child run is created during replay, and that fresh sequential calls still spawn.

All written TDD-style, each reproduced the duplicate-spawn before the fix.

### DIFF
--- a/packages/agentic/src/__tests__/workflow-runner.test.ts
+++ b/packages/agentic/src/__tests__/workflow-runner.test.ts
@@ -1341,6 +1341,191 @@ describe('WorkflowRunner', () => {
     }
   });
 
+  it('lets actions skip side effects via hasRecordedResult when replaying a resumed step', async () => {
+    const sideEffects: string[] = [];
+    const spawnAction = defineAction<
+      unknown,
+      { reply: string },
+      TestContext,
+      Settings
+    >({
+      name: 'spawn_then_wait_action',
+      inputSchema: z.any(),
+      outputSchema: z.object({ reply: z.string() }),
+      execute: async ({ context }) => {
+        // Deterministic replay re-runs this body; the guard must prevent the
+        // side effect from executing a second time.
+        if (!context.workflow.hasRecordedResult()) {
+          sideEffects.push('spawn');
+        }
+
+        const resumeData = (await context.workflow.suspend({
+          reason: 'awaiting_child_workflow',
+        })) as { reply: string };
+
+        return { reply: resumeData.reply };
+      },
+    });
+    const definition: WorkflowDefinition = {
+      defs: createTaskDefs({
+        wait_step: {
+          action: 'spawn_then_wait_action',
+          inputs: {},
+        },
+      }),
+      flow: [{ do: 'wait_step' }],
+      outputs: { reply: '=$output.wait_step.reply' },
+    };
+    const compiled = compileWorkflow(definition, {
+      actions: { spawn_then_wait_action: spawnAction },
+    });
+    const runner = new WorkflowRunner(compiled);
+    const startResult = await runner.start({
+      inputData: {},
+      context: new TestContext({}),
+    });
+
+    expect(startResult.status).toBe('suspended');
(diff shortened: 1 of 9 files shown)


yassinedorbozgithub (approved):
LGTM

PR #2121, opened 2026-07-12
fix: translation refresh pool exhaustion &  in nested expressions

marrouchi:
## Problem

Mentioned in issue #2119 

### 1. POST /api/translation/refresh self-deadlocks the DB pool beyond ~7 strings

The refresh endpoint fired all findOneOrCreate calls in parallel via Promise.all. Each write held a pool connection while its hook:translation:* event triggered a full resetI18nTranslations() (→ findAll), which needed yet another connection — exhausting the pool and deadlocking the request.

### 2. $t() broken inside array/object input literals

Nested = expressions found inside literal values were recompiled at evaluation time without the registered JSONata functions, so $t('...') failed anywhere but top-level expressions.

### API
- Moved the refresh logic from the controller into TranslationService.refreshWorkflowTranslations().
- Writes now run sequentially, and the event-driven i18n refresh is deferred (via an i18nRefreshDeferrals counter) to a single reload at the end of the batch — safe under concurrent refresh requests.
- Kept the idiomatic wildcard `@OnEvent('hook:translation:*')` handler (consistent with role, menu, setting, language services) so postUpdateMany and future hooks stay covered.
- Replaced the O(n²) indexOf dedupe with a Set; typed the controller's refresh() as Promise<DeleteResult>.

### Agentic Lib
CompiledValue now carries the jsonataFunctions registry, and resolveNestedExpressions threads it through when recompiling nested = strings — so custom functions like $t are registered on every nested expression, at any depth.

### DIFF
--- a/packages/agentic/src/__tests__/workflow-values.test.ts
+++ b/packages/agentic/src/__tests__/workflow-values.test.ts
@@ -92,6 +92,32 @@ describe('workflow values', () => {
     });
   });
 
+  it('registers custom functions on nested expressions inside literal values', async () => {
+    const translate = jest.fn((key: string) => `translated:${key}`);
+    const mapping = {
+      payload: compileValue(
+        {
+          title: "=$t('Title')",
+          actions: ['=$t("Confirm")', { label: "=$t('Cancel')" }],
+        },
+        { jsonataFunctions: { t: translate } },
+      ),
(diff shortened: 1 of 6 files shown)


yassinedorbozgithub (approved):
LGTM

PR #2125, opened 2026-07-14
Feat/enhance workflow validation ux

marrouchi:
## What changed?

When a workflow referenced an action missing from the server catalog (e.g. uninstalled extension), the graph spun forever with no explanation. This reworks definition-error handling end to end around a single structured issue model.

### Changes

- **agentic**: `validateWorkflow` / `validateAndResolveBindings` now return structured issues (`code`, `message`, yaml `path`, `actionName`/`taskId`) instead of flat `errors: string[]`; consumers derive strings via the new `issueMessages()` helper.
- **frontend state**: `useWorkflowDefinitionState` validates once and compiles only when valid, exposing `definitionStatus` (`loading | empty | invalid | ready`) + localized `definitionIssues` through the workflow context. Loading is gated on the catalog queries' real `isSuccess` flags (a failing catalog fetch now surfaces an error instead of an infinite spinner). Redundant detection paths removed (`extractActionNamesFromYaml`, `getDefinition`, duplicate task-reference collector).
- **UX**: one issue model drives four surfaces — in-graph error panel with an "Open YAML editor" CTA, error badge on the drawer's YAML toggle, the YAML editor alert, and issue-driven Monaco markers. While typing invalid YAML the last valid graph stays visible (dimmed, non-interactive) under the panel.
- **graph**: error-panel styles moved to CSS variables (`--workflow-error-*`) with light/dark theming; new `onOpenYamlEditor` host callback.
- **API**: exception messages derived from issues (unchanged wire text); MCP `hexabot_workflow_yaml_validate` now returns structured `{ valid, issues }` instead of `{ valid, errors }`.


yassinedorbozgithub (commented):
In this case the Yaml Editor has a validation error but the bubble don't indicate the error number 

Image

yassinedorbozgithub (approved):
LGTM

PR #2126, opened 2026-07-14
feat(frontend): add undo/redo buttons

marrouchi:
## What changed?

Undo/Redo in the workflow title bar, navigating the existing version chain by moving the workflow.currentVersion pointer — no new version rows are created, and no backend changes were needed.

## How it works

- **Undo** — if there are unsaved local edits, the first click reverts the canvas to the last saved version (and cancels the pending 4 s auto-commit). Otherwise it PATCHes currentVersion back to the version's parentVersion. The pointer change flows through the normalized query cache, which reloads the canvas automatically.
- **Redo** — walks a session-local stack of undone version ids forward. The stack is cleared whenever the pointer moves for any other reason (a save, an auto-commit, a restore from the versions panel, a publish, or switching workflows).
- **Buttons** sit in their own TitleBarCard between the title card and the save button, styled to match WorkflowSaveButton (lucide Undo2/Redo2, tooltips, disabled states). Disabled at the root version, while saving/navigating, and redo is disabled while dirty.


yassinedorbozgithub (approved):
Awesome feature 🚀 
LGTM ✅

PR #2136, opened 2026-07-16
fix(frontend): resolve yaml action inputs completions

yassinedorbozgithub:
## Screenshots
- After the fix
image

- Before the fix
image

## Summary
This PR fixes YAML autocompletion for action `inputs` by correctly detecting the enclosing action when editing inside a task's `inputs` block. Completions are now resolved only for task-level action inputs and no longer incorrectly trigger within workflow-level `inputs` declarations.

## Why
The previous detection logic could fail to identify the correct action context or confuse workflow-level and task-level `inputs`, resulting in missing or incorrect autocomplete suggestions. This improves the YAML editing experience and provides more accurate completions.

## How to review
* Review the logic responsible for resolving the enclosing action from the current cursor position.
* Verify that completions are only shown inside a task's `inputs` block.
* Confirm that workflow-level `inputs` schema declarations are ignored.

## Validation
* Verified autocompletion for multiple action types inside task `inputs`.
* Confirmed no completions are shown for workflow-level `inputs`.
* Tested nested tasks and different indentation levels to ensure the correct action context is resolved.

### DIFF
--- a/packages/frontend/src/components/visual-editor/v4/components/yaml-editor/completion.test.ts
+++ b/packages/frontend/src/components/visual-editor/v4/components/yaml-editor/completion.test.ts
@@ -7,6 +7,8 @@
 import type { Monaco } from "@monaco-editor/react";
 import { describe, expect, it, vi } from "vitest";
 
+import type { IAction } from "@/types/action.types";
+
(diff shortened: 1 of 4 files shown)


yassinedorbozgithub:
Closed in favor of #2139

PR #2172, opened 2026-07-27
feat(api): add sqlite-vector rag helper

yassinedorbozgithub:
## Summary
Adds `sqlite-vector`, a new RAG helper that brings embedding-based semantic search to SQLite deployments via the [`sqlite-vec`](https://github.com/asg017/sqlite-vec) extension. It is the SQLite counterpart to the existing `pgvector` helper — the two gate on the database driver, so exactly one is ever selectable on a given deployment. Also lifts the shared `chunkSearchText` implementation onto `BaseRagHelper`, removing a duplicated copy from `pgvector`.

## Why
Until now only PostgreSQL had an embedding-backed RAG helper. SQLite deployments fell back to `fulltext-search` (lexical only), so they got no semantic retrieval at all. This closes that gap without asking anyone to run Postgres: embeddings live in the existing SQLite file, with no extra service.

The chunker move is a correctness concern, not just tidiness. Chunk window boundaries are baked into every stored vector, so two embedding helpers that chunked differently would produce indexes that cannot be compared or migrated between. Having one implementation on the base class makes that structural rather than a convention two files happen to share.

## How to review
**Start with `sqlite-vector.provisioning.ts`** — the DDL and the two enqueue triggers are the genuinely new, dialect-specific part. Everything else is closely modelled on the proven `pgvector` helper.

Decisions worth a second opinion:

- **BLOBs, not `vec0` virtual tables.** Embeddings are `vec_f32` BLOBs in an ordinary table. `vec0` requires a fixed declared dimension, which the profile hash deliberately keeps fluid across model changes, and it cannot participate in the foreign keys that cascade content deletion. Search is an exact cosine scan either way, so there is no recall or index-quality tradeoff.
- **Where `sqlite-vec` is loaded** (`sqlite-vector.store.ts`) — lazily, into the  connection TypeORM already owns, rather than through the DataSource `prepareDatabase` hook. Keeps core DB config free of helper-specifi knowledge; a deployment that never selects this helper never loads it.
- **SQLite-specific query shapes.** No `FOR UPDATE SKIP LOCKED` (SQLite serializes writers, so `UPDATE … RETURNING` is already atomic); ISO-8601 TEXT timestamps produced by the process; retry backoff computed in JS rather than `make_interval`. Each is commented where it differs from the Postgres original.
- **No migration.** Tables are provisioned lazily and idempotently on first use.
- **The one `pgvector` change** is a two-line diff swapping an import for the inherited `this.chunkSearchText(...)`.

## Validation
- **Real-database integration suite** — 14 tests against an actual SQLite file covering provisioning idempotency, trigger-driven enqueueing on insert/`searchText`/`status` (and *no* enqueue on unrelated or no-op writes), cosine ranking, best-chunk-per-document, profile rollover, stale-claim rejection, retry backoff, lease behaviour, and FK cascade on content delete. It needs no external service, so it runs in normal CI; it self-skips if the `sqlite-vec` binary cannot load on the host.
- `sqlite-vector`: 47 tests passing across 4 suites.
- `pgvector`: 33 passing, unaffected by the chunker move (integration suite remains env-gated behind `TEST_PGVECTOR_DATABASE_URL`; its store was not touched).
- `helper/lib`: 7 passing, including new `base-rag-helper.spec.ts` coverage for the relocated chunker — the four original assertions plus one that two helper instances chunk identically.
- **Full API suite: 1308 passing, 0 failing.** Typecheck and lint clean.
- **Build verified** — `dist` regenerated from scratch; helper and settings land where the discovery globs expect them, i18n assets are copied, and specs are excluded.

### Known limitation
The API Docker image is Alpine (musl) and `sqlite-vec`'s prebuilt binaries are glibc-only. If the extension fails to load there, the helper still registers but reports a typed `RagHelperUnavailableError` with an actionable message on first use rather than silently returning no hits. Making that path work would mean adding `gcompat` to the image — out of scope here.


packages/api/src/extensions/helpers/sqlite-vector/index.helper.ts
```diff
@@ -0,0 +1,728 @@
...
+type SqliteVectorSettings = {
+  embedding_provider: string;
+  embedding_model: string;
+  embedding_api_key: string;
+  embedding_base_url: string;
+  embedding_dimensions: number;
+  chunk_size: number;
+  chunk_overlap: number;
+  index_only_active_content: boolean;
+};
```
marrouchi: Can we infer type from the zod schema ?

packages/api/src/extensions/helpers/sqlite-vector/index.helper.ts
```diff
@@ -0,0 +1,728 @@
...
+ * serving retrieval until the replacement is complete.
+ */
+@Injectable()
+export default class SqliteVectorRagHelper
+  extends BaseRagHelper<typeof SQLITE_VECTOR_RAG_HELPER_NAME>
+  implements OnApplicationBootstrap, OnApplicationShutdown
+{
+  private readonly store: SqliteVectorStore;
+
+  private readonly workerId = randomUUID();
```
marrouchi: Since, using sqlite, we won't have any parallel scaling, I don't think having a worker is a good idea, unless I'm missing something.

packages/api/src/helper/lib/base-rag-helper.ts
```diff
@@ -60,6 +66,72 @@ export abstract class BaseRagHelper<
...
+   * Lives here so every embedding-backed helper chunks identically: the window
+   * boundaries are baked into each stored vector, so two helpers that disagreed
+   * about them would produce indexes that cannot be compared or migrated
+   * between. Determinism matters for the same reason — re-chunking unchanged
+   * text must yield byte-identical windows, or every reconciliation pass would
+   * consider the whole corpus stale.
+   *
+   * @throws RangeError when the chunk size or overlap is not usable.
+   */
+  protected chunkSearchText(
```
marrouchi: Can we refactor pgvector to use the same chunking method ? Or do we need them to be different ?

marrouchi (changes requested):

packages/api/src/helper/lib/base-rag-helper.ts
yassinedorbozgithub: Happy to do that :technologist:

packages/api/src/extensions/helpers/sqlite-vector/index.helper.ts
yassinedorbozgithub: 100% i will fix the logic :technologist:

packages/api/src/extensions/helpers/sqlite-vector/index.helper.ts
yassinedorbozgithub: Sure

PR #2179, opened 2026-07-30
feat(api): extract pgvector helper to a seperate repo/pkg

marrouchi:
The PostgreSQL `pgvector` semantic RAG helper is now maintained and published from a separate repo as the optional `hexabot-helper-pgvector` package. This PR removes it from the monorepo and de-couples the core from it.

### Changes
- **Removed** the pgvector helper source (`extensions/helpers/pgvector/`) and all of its monorepo wiring: `pnpm-workspace.yaml`, `bump-version.sh`, `npm-publish.yml` / `npm-promote.yml`, and the CI integration step.
- **Slimmed the v3.4.0 migration** to core concerns only — it drops the legacy LlamaIndex footprint (searchText btree index + SQLite `content_chunks` / `content_embeddings` remnants), provisions lexical `fulltext-search`, and defaults `default_rag_helper` to `fulltext-search`. All pgvector provisioning and the legacy `rag_settings` migration now live in the external package (runtime self-heal +  bootstrap).
- **Exposed the public API** external RAG helpers extend: `BaseRagEmbeddingHelper` and the `content-search.store` utilities are now re-exported from `@hexabot-ai/api`.
- Kept a core PostgreSQL integration suite for the btree-index removal and lexical provisioning (`1784815200000-v-3-4-0.integration.spec.ts`).
- Switched the CI service and `docker-compose` DB image from `pgvector/pgvector` to plain `postgres:16` (the core no longer needs the `vector` extension).

### Notes
- The helper is still auto-discovered when installed (`node_modules/hexabot-helper-*/**/*.helper.js`) — no core wiring required.
- ⚠️ Deployments that install `hexabot-helper-pgvector` must point at a Postgres with the `vector` extension (e.g. the `pgvector/pgvector` image); the default compose stack now ships plain `postgres:16`.


.mcp.json
yassinedorbozgithub, line 1: I think this file was accidentally included

yassinedorbozgithub (approved):
LGTM 🚀

PR #2190, opened 2026-08-01
fix(settings): require storage helper selection

### LINKED ISSUE #1961: [Issue] Settings default storage helper need to be required
## Problem ##
We can remove the default storage helper from the settings UI.
This will hide all the attachments cross the pages (profile picture, step attachments, contents, message attachments)
 End user is not informed or guided in this situation.  

## Screenshots ## 
Image

Image

Image

Image

mikemikimike:
## Summary

- Remove the implicit `local-storage` default from the `default_storage_helper` setting so the settings UI requires an explicit helper selection.
- Keep seeding defaults for optional settings while allowing required settings to remain unset until configured.
- Add regression coverage for the JSON Schema and seed behavior.

Closes #1961

## Validation

- API lint ✅
- API typecheck ✅
- Targeted settings/Zod tests: **6 passed** ✅
- Settings test suite: **42 passed** ✅
- Full API Jest run: **1227 passed, 3 failed, 3 skipped**. The 3 failures are existing Windows path-separator assertions in migration and attachment tests; unrelated to this change.

The repository Husky hook could not run its test command on Windows because it uses POSIX syntax (`NODE_OPTIONS=...`). The equivalent Jest command was run directly and produced the results above. Local Node is 22.16.0 while the repository declares Node 24.17+; pnpm reports this as an engine warning.

### DIFF
--- a/packages/api/src/setting/default.settings.ts
+++ b/packages/api/src/setting/default.settings.ts
@@ -21,19 +21,16 @@ export const globalSettingsSchema = z
         'Provide the license key associated with your subscription. Learn more about available plans at https://hexabot.ai/pricing#pricing.',
       'ui:widget': 'password',
     }),
-    default_storage_helper: z
-      .string()
-      .default('local-storage')
-      .meta({
-        title: 'Default storage helper',
-        description: 'Helper used to persist workflow data by default.',
-        'ui:widget': 'AutoCompleteWidget',
-        'ui:options': {
-          entity: 'StorageHelper',
-          valueKey: 'name',
-          labelKey: 'name',
-        },
-      }),
+    default_storage_helper: z.string().meta({
+      title: 'Default storage helper',
+      description: 'Helper used to persist workflow data by default.',
+      'ui:widget': 'AutoCompleteWidget',
+      'ui:options': {
+        entity: 'StorageHelper',
+        valueKey: 'name',
+        labelKey: 'name',
+      },
+    }),
     default_rag_helper: z
       .string()
       .default('fulltext-search')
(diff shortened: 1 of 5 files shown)


yassinedorbozgithub (changes requested):
Hi @mikemikimike, thank you for your contribution.
I tested the fix you were working on, but the expected behavior has not been achieved yet.

I will share how we can reproduce the bug : 
1) Open the settings page 
http://localhost:8080/settings
2) Remove the default selected storage helper
image</img>
3) Open the profile page
http://localhost:8080/profile
4) Try to upload a profile picture
image</img>

mikemikimike:
Thanks for the reproduction. The missing case was that z.string() accepts the empty string.

Updated in commit 09593c749:
- default_storage_helper is now z.string().min(1), so clearing the AutoComplete field produces a validation error and cannot persist an empty helper name.
- Added a regression assertion that an empty helper value is rejected.
- The setting remains required in the generated JSON Schema and has no implicit default.

Validation: API lint passed; API typecheck passed; settings/Zod tests 45 passed; full API Jest 1227 passed, 3 skipped, with 2 pre-existing Windows path-separator suites failing in migration and attachment path assertions.

mikemikimike:
Browser validation completed locally with the API on port 3000 and the frontend temporarily on 8081:

- Logged in successfully with the local seeded admin account.
- Opened `/settings`: `Default storage helper` remains empty and required; it is no longer implicitly populated with `local-storage`.
- Opened `/profile`: the profile page and avatar upload control load normally.
- The browser connector allowed opening the file chooser but blocked injecting a local test image, so the final file-upload request could not be completed through browser automation.

The API and frontend processes were stopped after the check; no repository configuration changes were left behind.

mikemikimike:
Closing this PR as requested.

PR #2194, opened 2026-08-03
fix(api): support structured outputs for OpenAI-compatible providers

### LINKED ISSUE #2192: openai-compatible provider silently drops the JSON schema from structured generation
### What happens

`AiBaseAction.buildProviderInitOptions` returns `{ apiKey, baseURL, organization }` and nothing else, so `createOpenAICompatible` is called without `supportsStructuredOutputs`.

`@ai-sdk/openai-compatible` defaults that to `false`, and when it is false the provider **omits the JSON schema from `response_format` entirely**. Structured generation against an openai-compatible endpoint therefore depends on the model volunteering valid JSON rather than the endpoint enforcing it.

### Why it matters

It works most of the time, which is what makes it expensive. The failure is an unparseable response some fraction of the time rather than a clear error, so it reads as model flakiness rather than a missing request parameter.

We hit this on OpenRouter with `google/gemini-2.5-flash` — structured turns were intermittently returning prose.

### Why it can't be worked around

`aiModelBindingSchema` is a `z.strictObject`, so passing `supports_structured_outputs` through the model binding is rejected. There is no configuration path to turn it on.

### Suggested fix

Add `supports_structured_outputs` as an optional model-binding setting, shown for `litellm` / `openai-compatible`, and thread it into `buildProviderInitOptions`.

Leaving it undefined when unset keeps the provider's own default, so nothing changes for existing deployments — it only makes the capability reachable. Many openai-compatible endpoints do support `response_format: json_schema`; hardcoding `true` would be wrong for the ones that don't, which is why this is opt-in rather than a changed default.

### Versions

Reproduced on `@hexabot-ai/api` 3.3.3, still present in 3.4.0.

I have a branch with the change and tests covering the three states (unset, enabled, explicitly disabled) — happy to open it as a PR if the approach looks right.

yassinedorbozgithub:
## Screenshot
image

## Summary
Added an optional `supports_structured_outputs` model-binding setting for LiteLLM and OpenAI-compatible providers. The setting is forwarded to the AI SDK as `supportsStructuredOutputs`.

## Why
OpenAI-compatible providers default structured-output support to `false`, causing JSON schemas to be omitted from requests. This makes structured generation unreliable for compatible endpoints that support `response_format: json_schema` and resolves #2192.

## How to review
Focus on:

- The new optional field in `model.binding.ts`.
- The snake_case-to-camelCase mapping in `AiBaseAction.buildProviderInitOptions`.
- Coverage of enabled, disabled, and undefined configurations in `ai-base.action.spec.ts`.

## Validation
- Targeted AI base-action tests: 41 passed.
- API TypeScript typecheck passed.
- API lint passed.
- API production build passed.
- `git diff --check` passed.

### DIFF
--- a/packages/api/src/extensions/actions/ai/model.binding.ts
+++ b/packages/api/src/extensions/actions/ai/model.binding.ts
@@ -60,6 +60,21 @@ export const aiModelBindingSchema = z.strictObject({
         hideUntilAdded: true,
       },
     }),
+  supports_structured_outputs: z
+    .boolean()
+    .optional()
+    .meta({
+      title: 'Supports Structured Outputs',
+      description:
+        'Enable for endpoints that support response_format: json_schema. This sends the full JSON schema for structured generation. Leave unset to preserve the provider default.',
+      'ui:options': {
+        showWhen: {
+          field: 'provider',
+          in: ['litellm', 'openai-compatible'],
+        },
+        hideUntilAdded: true,
+      },
+    }),
 });
 
 declare global {
--- a/packages/api/src/extensions/actions/ai/model.binding.ts
+++ b/packages/api/src/extensions/actions/ai/model.binding.ts
@@ -70,7 +70,7 @@ export const aiModelBindingSchema = z.strictObject({
       'ui:options': {
         showWhen: {
           field: 'provider',
-          in: ['litellm', 'openai-compatible'],
+          in: ['gateway', 'litellm', 'openai-compatible'],
         },
         hideUntilAdded: true,
       },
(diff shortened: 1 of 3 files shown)


packages/api/src/extensions/actions/ai/model.binding.ts
```diff
@@ -60,6 +60,21 @@ export const aiModelBindingSchema = z.strictObject({
...
+    .boolean()
+    .optional()
+    .meta({
+      title: 'Supports Structured Outputs',
+      description:
+        'Enable for endpoints that support response_format: json_schema. This sends the full JSON schema for structured generation. Leave unset to preserve the provider default.',
+      'ui:options': {
+        showWhen: {
+          field: 'provider',
+          in: ['litellm', 'openai-compatible'],
```
marrouchi: Do you need this when the provider is "gateway" ?

marrouchi (changes requested):

packages/api/src/extensions/actions/ai/model.binding.ts
yassinedorbozgithub: Yes, gateway need it as well

